PR #32, opened 2019-12-16
global: Use libelogind when libsystemd is not available

### LINKED ISSUE #17: Consider making dependency on systemd optional
One hard requirement for even considering usage of `sid` within `udisks` is support for non-`systemd` Linux distributions. Systems running either forks like `eudev` or ripped-out and heavily patched `systemd-udev` from the `systemd` tree (nasty, but necessary).

Technically certain `systemd` API got reimplemented by independent projects, e.g. `elogind`, however most of the `systemd` core init daemon API is not available.

We simply can't afford limiting `udisks` availability to certain init systems, that would be too big regression given the current widespread of `udisks` among desktop and embedded systems. Maintaining two code paths keeping `udev` supported makes no sense either as the daemon is largely async and needs to be fine-tuned to prevent race conditions, etc. I cannot imagine continuous testing and support for both options.

Cc: @vojtechtrefny, @vpodzime

tbzatek:
`Elogind` ships with the `libelogind` library that is built upon `libsystemd` sources. Symbol names, header file naming and most of its original functionality is preserved.

https://github.com/elogind/elogind

Closes #17

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -86,7 +86,9 @@ AC_TYPE_UINT64_T
 AC_FUNC_ERROR_AT_LINE
 AC_FUNC_MALLOC
 
-PKG_CHECK_MODULES(SYSTEMD, libsystemd >= 221) # event loop supported in libsystemd version >= 221
+# event loop supported in libsystemd version >= 221
+PKG_CHECK_MODULES(SYSTEMD, libsystemd >= 221, [], [
+    PKG_CHECK_MODULES(ELOGIND, [libelogind >= 221])])
 
 PKG_CHECK_MODULES(UDEV, libudev >= 174) # support for device tagging added in libudev version >= 174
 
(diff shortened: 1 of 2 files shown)


tbzatek:
Quick and dirty patch. Nearly untested. It compiles, executes and reacts on `usid` commands.
```
$ usid version
USID_PROTOCOL=1
USID_MAJOR=0
USID_MINOR=0
USID_RELEASE=2
SID_PROTOCOL=1
SID_MAJOR=0
SID_MINOR=0
SID_RELEASE=2
$ usid active
USID_BRIDGE_STATUS=active
```

I'm getting lots of errors like
```
<command/20543/scan> scan-next phase failed.
<command/20543/scan> Failed to execute command
```
though they might be unrelated to this change.

tbzatek:
```
# sid -f -v
^C<sid> Event loop failed.: Unknown error 1419826320.
```

prajnoha:
> I'm getting lots of errors like
> 
> ```
> <command/20543/scan> scan-next phase failed.
> <command/20543/scan> Failed to execute command
> ```
> 
> though they might be unrelated to this change.

Yes, that's expected for now - this is because of missing modules to handle specific device types (I should probably change/add better message here so it's more clear what's wrong).

PR #38, opened 2020-01-21
fix #18: add initial rpm spec file to pack SID.

### LINKED ISSUE #18: add RPM spec file
We need a spec file to be able to pack SID as rpm package

grinrag:
The RPM spec file to build a package.
To test:
```sh
dnf copr enable galt/sid-project
```
```sh
dnf install sid
```

### DIFF
--- a/rpm/sid.spec
+++ b/rpm/sid.spec
@@ -1,0 +1,60 @@
+Name:           sid
+Version:        0.0.2
+Release:        mvp.1%{?dist}
+Summary:        Minimum viable product for SID project
+
+License:        GPLv2.0
+URL:            http://sid-project.github.io
+Source0:        https://github.com/sid-project/%{name}-mvp/archive/%{version}.tar.gz
+
+BuildArch:      x86_64
+
+BuildRequires:  systemd-devel >= 221
+BuildRequires:  libuuid-devel
+BuildRequires:  libblkid-devel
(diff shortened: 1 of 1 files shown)


rpm/sid.spec
```diff
@@ -0,0 +1,60 @@
...
+associated actions for well-defined triggers, including activation and deactivation of devices
+and their layers in the stack.
+
+%prep
+%setup -q -c %{name}-%{version}
+mv %{name}-mvp-%{version}/* .
+
+%build
+./autogen.sh
+./configure --disable-mod-multipath_component CC=gcc CLFAGS="-g0 -O0 -Wall"
```
prajnoha: ...a little typo here: CFLAGS, not CLFAGS

grinrag: Interestingly, I don't see the error with multiple definitions in the Copr. Here is the log:
https://copr-be.cloud.fedoraproject.org/results/galt/sid-project/fedora-rawhide-x86_64/01170699-sid/builder-live.log.gz

prajnoha: Hmm, I even tried the version from the copr (`glibc-common              x86_64 2.30.9000-29.fc32 `), but still getting those `ld` errors on linkage. Anyway, let's have this in, this seems to be something specific to my machine - when we find out, we'll fix that.

PR #40, opened 2020-02-03
rpm: disable debugsource subpackage building.

grinrag:
The Copr build is failed because of missing file for debugsource package.
I disabled the build of debugsource.

### DIFF
--- a/rpm/sid.spec
+++ b/rpm/sid.spec
@@ -35,7 +35,7 @@ mv %{name}-mvp-%{version}/* .
 
 %build
 ./autogen.sh
-./configure --disable-mod-multipath_component CC=gcc CFLAGS="-g0 -O0 -Wall"
+./configure --disable-mod-multipath_component CC=gcc
 make
 
 %install


prajnoha:
Aha! So we don't get the debuginfo then for some reason... I've looked a bit more closely and I've noticed that we have `+ ./configure --disable-mod-multipath_component CC=gcc 'CFLAGS=-g0 -O0 -Wall'` in Copr build logs. The `-g0` disables the debuginfo for binaries so that's probably why Copr is not able to get them - so try passing just `-g` (without `0`).

Now, I think we should also pass global `CFLAGS` as set by RPM macros (see `/usr/lib/rpm` content and the all the `optflags` there which have `-O2 -g` defined by default). Then we should probably redefine the ones we really need to (e.g. the -O2 to -O0 if we want to have better debug info).

Hmm, try also including these 62b9500758ef55b9b1459240321e2d4757882cfb and afb5c8ec67a17aaa3101be262a63ed183cd92827.

These should resolve another issue with:

They should help with these errors on rawhide (just pasting one example error, but there is one issued for each SID resource we have):

```usr/bin/ld: .libs/libsidresource_la-aggregate.o:(.data.rel.ro.local+0x0): multiple definition of `sid_resource_type_aggregate'; .libs/libsidresource_la-resource.o:(.rodata+0x0): first defined here```

Seems like that the GCC version on rawhide is a bit more stricter about how we declare and define resource types among multiple files in the source code.

prajnoha:
> Now, I think we should also pass global `CFLAGS` as set by RPM macros (see `/usr/lib/rpm` content and the all the `optflags` there which have `-O2 -g` defined by default). Then we should probably redefine the ones we really need to (e.g. the -O2 to -O0 if we want to have better debug info).

(The reason for including the globally defined `CFLAGS` from RPM macros is that there could be a policy in the distro that mandates use of certain build options. We'd need to check the Fedora guidelines here probably if there's something we need to follow within distribution.)

PR #41, opened 2020-04-28
Fixes

bmarzins:
Here are some issues I ran across while reading the SID code to come up to speed. This pull request also contains commits add a "tests" directory and unit tests using cmocka to verify some of these issues and my fixes. The unit tests can be run with "make check" from the tests directory.  cmocka's reporting needs a bit of work to make it play nicer with autotools. Right now you have to look at the
.log files to see what went wrong. If you don't want to pull in the unit tests, that's fine. Just let me know, and I'll pull them into a separate branch, and just include the fixes in the pull request.

### DIFF
--- a/src/iface/service-link-iface.c
+++ b/src/iface/service-link-iface.c
@@ -133,7 +133,7 @@ int service_link_group_remove_member(struct service_link_group *slg, struct serv
 	return 0;
 }
 
-static const char *_get_arg_line(const char *str, const char *key_eq, size_t *size)
+static const char *_get_arg_value(const char *str, const char *key_eq, size_t *size)
 {
 	const char *str_end;
 	const char *line_end;
@@ -147,10 +147,9 @@ static const char *_get_arg_line(const char *str, const char *key_eq, size_t *si
 	for (str_end = str + strlen(str); str < str_end; str = line_end + 1) {
 		line_end = strchr(str, '\n') ? : str_end;
 
-		if (!strcmp(key_eq, str)) {
+		if (!strncmp(key_eq, str, strlen(key_eq))) {
 			/* get the value and its size */
 			str += strlen(key_eq);
-
 			*size = line_end - str;
 			return str;
 		}
@@ -168,7 +167,7 @@ int _do_service_link_notify(struct service_link *sl, struct service_link_group *
                             service_link_notification_t notification, const char *fmt, va_list ap)
 {
 	struct buffer *buf = NULL, *fmt_buf = NULL;
-	const char *arg_str, *arg_line;
+	const char *arg_str, *arg_value;
 	size_t size;
 	int unset = 0;
 	int iter_r, r = 0;
@@ -197,19 +196,31 @@ int _do_service_link_notify(struct service_link *sl, struct service_link_group *
 		unset = 1;
 
 	if (notification & SERVICE_NOTIFICATION_STATUS) {
-		if ((arg_line = _get_arg_line(arg_str, SERVICE_KEY_STATUS EQ, &size)))
-			if (!buffer_add(buf, (void *) arg_line, size)) {
+		if ((arg_value = _get_arg_value(arg_str, SERVICE_KEY_STATUS EQ, &size))) {
+			if (!buffer_fmt_add(buf, SERVICE_KEY_STATUS EQ "%.*s\n",
+			                    size, arg_value)) {
 				r = -ENOMEM;
 				goto out;
 			}
+			if (buffer_rewind(buf, 1, BUFFER_POS_REL) < 0) {
+				r = -errno;
+				goto out;
+			}
+		}
 	}
 
 	if (notification & SERVICE_NOTIFICATION_ERRNO) {
-		if ((arg_line = _get_arg_line(arg_str, SERVICE_KEY_ERRNO EQ, &size)))
-			if (!buffer_add(buf, (void *) arg_line, size)) {
+		if ((arg_value = _get_arg_value(arg_str, SERVICE_KEY_ERRNO EQ, &size))) {
+			if (!buffer_fmt_add(buf, SERVICE_KEY_ERRNO EQ "%.*s\n",
+			                    size, arg_value)) {
 				r = -ENOMEM;
 				goto out;
 			}
+			if (buffer_rewind(buf, 1, BUFFER_POS_REL) < 0) {
+				r = -errno;
+				goto out;
+			}
+		}
 	}
 
 	if (notification & SERVICE_NOTIFICATION_READY)
--- a/src/iface/service-link-iface.c
+++ b/src/iface/service-link-iface.c
@@ -24,15 +24,13 @@
 #include <stdlib.h>
 #include <string.h>
 
-#define SERVICE_KEY_READY     "READY"
-#define SERVICE_KEY_RELOADING "RELOADING"
-#define SERVICE_KEY_STOPPING  "STOPPING"
-#define SERVICE_KEY_WATCHDOG  "WATCHDOG"
+#define SERVICE_READY_LINE		"READY=1\n"
+#define SERVICE_RELOADING_LINE		"RELOADING=1\n"
+#define SERVICE_STOPPING_LINE		"STOPPING=1\n"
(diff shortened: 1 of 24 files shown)


bmarzins:
Oh, also the first patch, for multipath is using the new version of the libmpathvalid library that I recently posted to dm-devel.  It's still not accepted, and might end up changing again before it is, so if you want, I can pull that out for now, until the library API is finalized upsream.

bmarzins:
The multipath branch with the current libmpathvalid code (on top of the latest fedora code) is:
 [https://github.com/bmarzins/rh-multipath-tools/tree/mpathvalid-test](url)

bmarzins:
I have some other questions about the code. If you would like me to turn these into Issues instead, just let me know. While the code calls buffer_read() in a loop, it doesn't call buffer_write() in a loop. Since the on the daemon side, the socket is created nonblocking, is seems like the write could block.  I'm not sure if how the usid command will be used if it becomes a udev builtin, but it seems possible that it could receive a signal, and return EINTR, when trying to write.  Or is this not possible for a reason that I'm overlooking.

bmarzins:
In _passes_global_reservation_check()

https://github.com/sid-project/sid-mvp/blob/be11b4251612fb74559bcb8c8cea32e34029bfdb/src/resource/ubridge.c#L733-L739

we check if the there is a reservation flag, but the owner matches, and if so, we pass. But if that check fails, we fail.  This means that if the owner matches, but there is no reservation flag, we fail. Also, if the owner doesn't match, but there is no reservation flag, we fail. Is this correct?

bmarzins:
in _connection_cleanup()

https://github.com/sid-project/sid-mvp/blob/be11b4251612fb74559bcb8c8cea32e34029bfdb/src/resource/ubridge.c#L2988-L2996

I don't get the iter code. Aren't the children of conn_res already destroyed by sid_resource_destroy(conn_res)?

bmarzins:
Most of the time sid_ubridge_cmd_get_kv() is called, it's called with a key that starts with '#'
https://github.com/sid-project/sid-mvp/blob/be11b4251612fb74559bcb8c8cea32e34029bfdb/src/resource/ubridge.c#L93-L98
But looking at the check at the start of sid_ubridge_cmd_get_kv():
https://github.com/sid-project/sid-mvp/blob/be11b4251612fb74559bcb8c8cea32e34029bfdb/src/resource/ubridge.c#L901-L904
Won't all these calls fail. Am I missing something here?

bmarzins:
In _do_sid_ubridge_cmd_mod_reserve_kv(), if kv_store_res belongs to a worker, but unset is true you will still set the value. I get that you are removing the KV_MOD_RESERVED flag, but what if the key wasn't in the kv store at all, aren't you adding it. I'm still a bit fuzzy on the difference between what the worker stores, and what the main kv store stores, so maybe this is intentional, or maybe I'm missing something.

prajnoha:
Hey Ben, thanks a lot for looking through the code! Just looking at your patches now...

prajnoha:
>  This pull request also contains commits add a "tests" directory and unit tests using cmocka to verify some of these issues and my fixes. The unit tests can be run with "make check" from the tests directory. cmocka's reporting needs a bit of work to make it play nicer with autotools. Right now you have to look at the
> .log files to see what went wrong. If you don't want to pull in the unit tests, that's fine. Just let me know, and I'll pull them into a separate branch, and just include the fixes in the pull request.

Honestly, I haven't decided about unit testing framework yet, so let's go with the cmocka and we'll see if that is useful... I've been looking into CUnit a bit before, but someone told me that it's not in a good shape these days (but there's a fork somewhere on github). Anyway, I'm taking the cmocka - it looks easy to use and useful at glance.

src/iface/service-link-iface.c
```diff
@@ -144,13 +144,14 @@ static const char *_get_arg_line(const char *str, const char *key_eq, size_t *si
...
-			str += strlen(key_eq);
-
+	for (str_end = str + strlen(str); str < str_end; str = line_end) {
+		line_end = strchr(str, '\n');
+		if (line_end)
+			line_end++;
+		else
+			return NULL;
+
+		if (!strncmp(key_eq, str, strlen(key_eq))) {
```
prajnoha: I'd lean here to the way where we still allow the end of the overall arg string to be just \0, not \n\0, because it could be very easy to forget the \n at the end, mainly if it's used outside as a library function by others.

So I'd probably leave the existing loop and checks as they were (also returning just the value) and then let the _do_service_link_notify decorate the value with the proper key. This will also let us use a different key for the actual notification if needed (or different format than key=value pair). That "decoration" could all happen in _do_service_link_notify then (...or calling out specific functions centrally from there).

Of course, the strcmp is a bug, there needs to be the strncmp as you patched...

prajnoha:
> I have some other questions about the code. If you would like me to turn these into Issues instead, just let me know. 

(Yes, please do, so we have that on the issue list...)

> While the code calls buffer_read() in a loop, it doesn't call buffer_write() in a loop. Since the on the daemon side, the socket is created nonblocking, is seems like the write could block. I'm not sure if how the usid command will be used if it becomes a udev builtin, but it seems possible that it could receive a signal, and return EINTR, when trying to write. Or is this not possible for a reason that I'm overlooking.

Yes, you're right, we should add `buffer_write` to a loop too. The `buffer_write` call in `_cmd_handler` even has `(void)` check for return value - I'm planning to go over all these `(void)` return code checks and add proper ones. The `write` is actually the one which I should have done long time ago, I just forgot!

prajnoha:
> In _passes_global_reservation_check()
> 
> https://github.com/sid-project/sid-mvp/blob/be11b4251612fb74559bcb8c8cea32e34029bfdb/src/resource/ubridge.c#L733-L739
> 
> we check if the there is a reservation flag, but the owner matches, and if so, we pass. But if that check fails, we fail. This means that if the owner matches, but there is no reservation flag, we fail. Also, if the owner doesn't match, but there is no reservation flag, we fail. Is this correct?

Actually, there's a call to KV store right before the check:
```
     if (!(found = kv_store_get_value(cmd->kv_store_res, full_key, &value_size, &value_flags)))                                                                                                                                                             
                goto out;
```

The `full_key` there is set up in a way to find a record for the global reservation - the "global reservation record" is the one without any domain or ID set (so it's not bound to anything), but it only has the KV_MOD_RESERVED flag set (see also `struct kv_key_spec key_spec` at the beginning of the function).

So if we don't find such "global reservation" record in KV store, there's nothing to check and we simply "pass" right away. If the record is found, we check if it has the KV_MOD_RESERVED flag set and if yes, we check the owner - that's the check line you mention.

prajnoha:
> in _connection_cleanup()
> 
> https://github.com/sid-project/sid-mvp/blob/be11b4251612fb74559bcb8c8cea32e34029bfdb/src/resource/ubridge.c#L2988-L2996
> 
> I don't get the iter code. Aren't the children of conn_res already destroyed by sid_resource_destroy(conn_res)?

Yes, indeed, the iter + destroy here is superfluous. Probably a remnant of some old code...

prajnoha:
> Most of the time sid_ubridge_cmd_get_kv() is called, it's called with a key that starts with '#'
> https://github.com/sid-project/sid-mvp/blob/be11b4251612fb74559bcb8c8cea32e34029bfdb/src/resource/ubridge.c#L93-L98
> 
> 
> But looking at the check at the start of sid_ubridge_cmd_get_kv():
> https://github.com/sid-project/sid-mvp/blob/be11b4251612fb74559bcb8c8cea32e34029bfdb/src/resource/ubridge.c#L901-L904
> 
> 
> Won't all these calls fail. Am I missing something here?

Good catch again... the records with keys starting "#" are supposed to be "system records", so not directly accessible by API users (through the `sid_ubridge_cmd_get_kv` function).

BUT, this exact function is also called by some of SID's own internal fcode - so yes, we can't be checking for "#"  and fail if it's SID core itself calling this, only if it's called from outside by API users. We need to separate these two!

bmarzins:
I redid the commits "tests: more _do_service_link_notify() tests" and "iface: fix for _get_arg_line()" to make _do_service_link_notify() build the line and include the newline. Everything else is the same, except the two new commits about the issues we discussed above.

PR #44, opened 2020-07-23
Make multipath_component claim devices

bmarzins:
This updates the multipath component code to set the udev property to claim multipath devices.  It also updates the udev rules to make call usid on add events, and makes the multipath component respect the kernel cmdline multipath settings.

### DIFF
--- a/src/base/util.c
+++ b/src/base/util.c
@@ -24,6 +24,9 @@
 #include <stdlib.h>
 #include <string.h>
 #include <sys/stat.h>
+#include <unistd.h>
+#include <sys/stat.h>
+#include <fcntl.h>
 
 /*
  * Process-related utilities.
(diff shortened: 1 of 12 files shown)


bmarzins:
There are some design questions implicit in theses patches. For instance, should we move the kernel command line parsing out of the multipath module? Is the sid_ubridge_cmd_part_get_disk_kv() code (getting a parent device's DEVICE namespace values) the proper way to solve this issue. The idea was to recreate the udev ability to import parent values, but perhaps there is a better way to go about this in sid.

prajnoha:
(...sorry for the delay, I've been delved into the SID code recently, fixing a few issues I've found...)

> For instance, should we move the kernel command line parsing out of the multipath module?

Yup, I'd probably do this as part of libsidmisc probably (the util.h one) so others can make use of it too...

> Is the sid_ubridge_cmd_part_get_disk_kv() code (getting a parent device's DEVICE namespace values) the proper way to solve this issue. The idea was to recreate the udev ability to import parent values, but perhaps there is a better way to go about this in sid.

Hmm, good question. I think the ideal would be if we could generalize this for the stack and we'd be able to get (accessible/readable) information from the db if we have a path in the stack from device A to device B in general, not just whole dev - partition relation. Maybe that could even be a different kind of namespace (besides existing UDEV/DEVICE/MODULE/GLOBAL), like a namespace for "information visible within dev stack". It also depends on whether people would use this for something else than the whole dev - partition case :)

bmarzins:
This version also makes the unit tests work again.

I added util_cmdline_get_arg() to util.c, but any module that wants to read from the kernel command line will have to re-read /proc/cmdline. I wonder if it would make more sense for the sid daemon to read it in once during startup, since it never changes. The code is already split between _get_proc_cmdline() which reads the cmdline in, and util_cmdline_get_arg(), which searches it for the requested argument.

I left sid_ubridge_cmd_part_get_disk_kv() as is. I don't see how you would have a consistent namespace identifier for all the devices in a dev stack, since devices can belong to multiple dev stacks.

prajnoha:
Yes, exactly, I've been thinking about the same (parsing the kernel cmdline once) - I'll try to put that somewhere so it's also shared with worker processes. The sid_ubridge_cmd_part_get_disk_kv - OK, let's keep that one - if there's any need for that generic access to the stack in the future, we can revisit this... Otherwise, the patches look good, so I'll merge them. Thanks a lot!

PR #47, opened 2020-09-28
modules: ucmd: block: dm_mpath: use updated libmpathvalid API

bmarzins:
Sorry this took so long. It took longer to agree on all the upstream library changes than I thought. I've pushed a device-mapper-multipath package with the API changes to fedora [https://bodhi.fedoraproject.org/updates/FEDORA-2020-6934494270](https://bodhi.fedoraproject.org/updates/FEDORA-2020-6934494270). It's currently in testing, but I can push it to stable whenever, since it has positive karma.


prajnoha:
Great, thanks!

I'll prepare the SID package now, if we need to push the mpath package, I'll let you know, but I think it'll be already there in the stable at the time SID lands there (...there's some procedure first that SID needs to pass for the "new" package so that will take a bit of time).

Issue #22, opened 2019-11-15
Restrict access to SID socket


prajnoha:
With current code, SID uses unix socket in abstract namespace, not exposed on filesystem as a file. As such, there's no access control done so common users can also access the socket which is definitely not what we want for a service like SID.

We need either:

  - keep using the unix abstract socket, but enforce access controls by checking `SO_PEERCRED` on the socket,

  - or switch to filesystem-based unix socket where permissions are checked as usual.

Evaluate pros and cons of both methods and choose one.

prajnoha:
We're still using unix abstract socket that has the advantage of not requiring mounted filesystem to be available. Ben has patched the socket access per command in patch e263353388c60c720ae68991919f0eeffd70a477.

I'll keep this issue open to go through this once again and check thoroughly we have everything covered properly here...

PR #48, opened 2020-12-03
Add sidctl and cleanup issues found while writing it

bmarzins:
This adds sidctl, which can currently only issue the dump and version commands (I didn't remove the version command from usid, but I'm not sure if it belongs there if sidctl exists). On the server, the command works by building the dump response in the gen_buf, and then adding that buffer's memory to the res_buf. Since res_buf is a vector buffer, I needed some way to allocate the memory for the iovecs to point to, which would get cleaned up later. The gen_buf is a linear buffer with a large alloc_step, which is what we want for building the dump response. I had some other ideas for how to go about this, but they involved more changes to the general command handling code. If you have any issues with my solution, we can discuss the other possiblities. The rest of the patches are issues I noticed while working on this.  Again, for some of them I just picked one out of several possible solutions, so if you have a different way you'd like them solved, just say so. Also, the man page for sidctl still needs to be done.


prajnoha:
> This adds sidctl, which can currently only issue the dump and version commands
Great, thanks!

> (I didn't remove the version command from usid, but I'm not sure if it belongs there if sidctl exists).
I'd keep it there in both usid and sidctl just in case we'd like to version the tools separately for some reason...

> On the server, the command works by building the dump response in the gen_buf, and then adding that buffer's memory to the res_buf. Since res_buf is a vector buffer, I needed some way to allocate the memory for the iovecs to point to, which would get cleaned up later. The gen_buf is a linear buffer with a large alloc_step, which is what we want for building the dump response. I had some other ideas for how to go about this, but they involved more changes to the general command handling code. If you have any issues with my solution, we can discuss the other possiblities.

I've no objections... the gen_buf is there primarily for use cases where we need to do temporary allocations throughout the code and then rewind the buffer for another reuse (e.g. string catenations, temporary results...). Here in the dump command, we collect the dump and send it over to the client, then we destroy the command resource right away (there's nothing else to do in that "dump command") and hence the gen_buf gets destroyed too (the gen_buf is attached to the command resource). So I think we're OK here with this use.

What were the other ideas? Just curious....

> The rest of the patches are issues I noticed while working on this. Again, for some of them I just picked one out of several possible solutions, so if you have a different way you'd like them solved, just say so.

Looks fine, thanks! Mainly the checking for "root" on socket request which got out of my radar somehow - I even have issue #22 for that, so I can probably close that now. There, I was still considering whether to stay with the abstract unix socket or filesystem-based one, but I think we'd be better off with the abstract one as we don't need to think about where to place the filesystem-based one, even though this one would provide us the usual owner/group/mode settings... Both have some pros and cons, but I was always leaned to the abstract one, that's why I kept it in the end...

Also the vector buffer's partial writes and buffer_write_all - this was missing there!

> Also, the man page for sidctl still needs to be done.

Yeah... before release :)

bmarzins:
> > On the server, the command works by building the dump response in the gen_buf, and then adding that buffer's memory to the res_buf. Since res_buf is a vector buffer, I needed some way to allocate the memory for the iovecs to point to, which would get cleaned up later. The gen_buf is a linear buffer with a large alloc_step, which is what we want for building the dump response. I had some other ideas for how to go about this, but they involved more changes to the general command handling code. If you have any issues with my solution, we can discuss the other possiblities.
> 
> I've no objections... the gen_buf is there primarily for use cases where we need to do temporary allocations throughout the code and then rewind the buffer for another reuse (e.g. string catenations, temporary results...). Here in the dump command, we collect the dump and send it over to the client, then we destroy the command resource right away (there's nothing else to do in that "dump command") and hence the gen_buf gets destroyed too (the gen_buf is attached to the command resource). So I think we're OK here with this use.
> 
> What were the other ideas? Just curious....

The two other ideas I had were to either change the type of the res_buf depending on the command type or to have commands be able to signal that they will handle the rest of the work that _cmd_handler would normally do.

src/resource/ubridge.c
```diff
@@ -512,6 +512,59 @@ static const char *_get_iov_str(struct buffer *buf, bool unset, struct iovec *io
...
+	}
+
+	while ((value = kv_store_iter_next(iter, &size, &flags))) {
+		key = kv_store_iter_current_key(iter);
+		if (!strncmp(key, "U:", 2))
+			continue;
+		iov = _get_value_vector(flags, value, size, tmp_iov);
+		header.seqnum = KV_VALUE_SEQNUM(iov);
+		header.flags = KV_VALUE_FLAGS(iov);
+		header.data_count = (flags & KV_STORE_VALUE_VECTOR)? size - KV_VALUE_IDX_DATA : 1;
```
prajnoha, line 541: I think there's an issue when we use this for KV pairs where the value is blank (e.g. key reservations).

src/resource/ubridge.c
```diff
@@ -512,6 +512,59 @@ static const char *_get_iov_str(struct buffer *buf, bool unset, struct iovec *io
...
+			len = iov[KV_VALUE_IDX_OWNER + i].iov_len;
+			if (!buffer_add(buf, &len, sizeof(len), &r))
+				goto out;
+			if (!buffer_add(buf, data, len, &r))
+				goto out;
+		}
+	}
+	/* add zeroed header to signal the end of the dump */
+	memset(&header, 0, sizeof(header));
+	buffer_add(buf, &header, sizeof(header), &r);
```
prajnoha, line 560: This is probably not needed? We also get the size of the message on the other end (`sidctl`), so we should already know when to end without a need to check for the trailing zeroed header.

prajnoha:
> The two other ideas I had were to either change the type of the res_buf depending on the command type or to have commands be able to signal that they will handle the rest of the work that _cmd_handler would normally do.

Yes, I've had the same intention in my head after I looked more into the code due to the issues with blank KV values!

I realized that we could possibly even reuse the existing `_export_kv_store` function that we currently use to serialize and send db differences from worker to main process. We could make it conditional to send the whole db as well which would be suitable for whole db dump then.

The `_export_kv_store` uses `memfd_create` to create anonymous memory with an FD attached which we can send through the socket. So I'm now quite leaned to creating a `buffer` abstraction on top of `memfd` as an alternative backend to `malloc` we currently use for buffers. This could be then used for other communication in SID and its parts where appropriate.

I'm just trying to extend the buffer to include the memfd backend... and it looks feasible... Just completing it...

Then, as you say, we could simply use different buffer types based on commands we process. So for the dump, we'd use the memfd backed buffer - that would also save us copying the buffer content to the send buffer on `sid` side, sending it all through the socket, and then copying received data to receive buffer on `sidctl` side. With memfd, we'd just write the content once, send the FD, and read directly from the FD on the other side.

I'll push the buffer extension, then we can try to patch the `_export_kv_store` to use it and reuse it for the dump.

PR #50, opened 2021-01-11
Update signal handling for systemd conventions

trgill:
A systemd process is required to use sd_event_add_signal() rather than
signal() to get notifications of process signals.

Updated to use existing sid_resource_create_signal_event_source() to
handle SIGINT and SIGTERM.

Signed-off-by: Todd Gill <tgill@redhat.com>

### DIFF
--- a/src/daemon/sid.c
+++ b/src/daemon/sid.c
@@ -38,8 +38,6 @@
 
 #define KEY_VERBOSE "VERBOSE"
 
-static volatile sig_atomic_t _shutdown_requested = 0;
-
 static void _help(FILE *f)
 {
 	fprintf(f, "Usage: sid [options]\n"
@@ -60,34 +58,19 @@ static void _version(FILE *f)
 	        SID_COMPILATION_HOST, SID_COMPILER);
 }
 
-static void _shutdown_signal_handler(int sig __attribute__((unused)))
-{
-	_shutdown_requested = 1;
-}
 
 static void _become_daemon()
 {
-	pid_t pid = 0;
-	int child_status;
-	struct timeval tval;
 	int fd;
 
-	signal(SIGTERM, &_shutdown_signal_handler);
-
-	if ((pid = fork()) < 0) {
-		log_error_errno(LOG_PREFIX, errno, "Failed to fork daemon");
-		exit(EXIT_FAILURE);
-	}
-
-	if (pid > 0) {
-		log_debug(LOG_PREFIX, "Forked SID with pid=%d", pid);
-		while (!waitpid(pid, &child_status, WNOHANG) && !_shutdown_requested) {
-			tval.tv_sec = 1;
-			tval.tv_usec = 0;
-			select(0, NULL, NULL, NULL, &tval);
-		}
-
-		exit(_shutdown_requested ? EXIT_SUCCESS : WEXITSTATUS(child_status));
+	switch (fork()) {
+		case -1:
+			log_error_errno(LOG_PREFIX, errno, "Failed to fork daemon");
+			exit(EXIT_FAILURE);
+		case 0:
+			break;
+		default:
+			exit(EXIT_SUCCESS);
 	}
 
 	if (!setsid()) {
@@ -122,8 +105,6 @@ static void _become_daemon()
 
 	umask(SID_DEFAULT_UMASK);
 
-	if (kill(getppid(), SIGTERM) < 0)
-		log_error_errno(LOG_PREFIX, errno, "Failed to send SIGTERM signal to parent");
 }
 
 static sid_resource_service_link_def_t service_link_defs[] = {{
--- a/src/resource/resource.c
+++ b/src/resource/resource.c
@@ -886,8 +886,10 @@ int sid_resource_run_event_loop(sid_resource_t *res)
 
 int sid_resource_exit_event_loop(sid_resource_t *res)
 {
-	if (!res->sd_event_loop)
+	if (!res->sd_event_loop) {
+		log_debug(res->id, "sid_resource_exit_event_loop call with NULL event loop.");
 		return -ENOMEDIUM;
+	}
 
 	return sd_event_exit(res->sd_event_loop, 0);
 }
--- a/src/resource/sid.c
+++ b/src/resource/sid.c
@@ -24,6 +24,28 @@
 
 #define SID_NAME "sid"
 
+static int _sid_signal_handler(sid_resource_event_source_t *es, const struct signalfd_siginfo *si, void *data)
+{
+	switch (si->ssi_signo) {
+		case SIGINT:
+			break;
+		case SIGPIPE:
+			break;
+		case SIGHUP: /* TODO: Reload config on SIGHUP? */
+			break;
+		case SIGCHLD:
+			break;
+		default:
+			break;
+	};
+}
+
+static int _on_sid_signal_event(sid_resource_event_source_t *es, const struct signalfd_siginfo *si, void *arg)
+{
+	sid_resource_t *res = arg;
+	sid_resource_exit_event_loop(res);
+}
+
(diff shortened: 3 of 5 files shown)


trgill:
@prajnoha this needs more work. It is an improvement - but still not working properly.

The forking-and-exiting only makes sense if we start the daemon manually from a terminal and not with systemd, right?  I see the sid.service file starts it in the foreground.

Currently the parent hits the select() and gets the SIGTERM from the child and exits without handling the signal.  This leaves a defunct process until the child exits.

This is probably a low priority issue, right?  I picked it somewhat randomly.  I'll look through the other issues to see where I might be able to help.  Please let me know if you have any suggestions.

trgill:
@prajnoha I think this is ready for review.  Thanks.

src/daemon/sid.c
```diff
@@ -122,8 +104,6 @@ static void _become_daemon()
 
 	umask(SID_DEFAULT_UMASK);
 
-	if (kill(getppid(), SIGTERM) < 0)
```
prajnoha, line 125: There's only one implication of removal of this child to parent signalling...

When daemonizing, there's some little initialization done in the child process (after `fork` and before `kill` in the `_become_daemon` function) and if anything failed during this time, the child could still signal the parent about the failure and the parent could then return that status to the calling process (e.g. a command line from which SID is executed). However, the initialization is only `setsid`, `chdir` and a few `dup2` calls, so the signalling code is probably not worth for this. It would make sense if we had more complex initialization, like including reinstating the internal daemon state from permanent storage. So ok, let's dump it...

src/resource/resource.c
```diff
@@ -451,8 +451,28 @@ int sid_resource_create_io_event_source(sid_resource_t *res, sid_resource_event_
...
+			break;
+		case SIGPIPE:
+			break;
+		case SIGHUP: /* TODO: Reload config on SIGHUP? */
+			break;
+		case SIGCHLD:
+			break;
+		default:
+			break;
+	};
```
prajnoha, line 468: Well, I think we should move this out of `resource.c` to the caller. The `resource.c` is generic reusable code, the caller itself should register signals and possible handlers it needs. What if someone wants to define a resource and use resource API and doesn't want the SIGTERM to exit the event loop, but do something else. So for the `SIGTERM` (and other signals), we need to register a handler directly in the caller instead (which is `src/resource/sid.c` here).

Then if the caller wants to exit the event loop, it can use the `sid_resource_exit_event_loop` resource API function for that.

src/resource/sid.c
```diff
@@ -45,7 +48,10 @@ static int _init_sid(sid_resource_t *res, const void *kickstart_data, void **dat
 	}
 
 	if (sid_resource_create_signal_event_source(res, NULL, SIGTERM, NULL, 0, "sigterm", NULL) < 0 ||
-	    sid_resource_create_signal_event_source(res, NULL, SIGINT, NULL, 0, "sigint", NULL) < 0) {
+	    sid_resource_create_signal_event_source(res, NULL, SIGINT, NULL, 0, "sigint", NULL) < 0 ||
+	    sid_resource_create_signal_event_source(res, NULL, SIGPIPE, NULL, 0, "sigpipe", NULL) < 0 ||
+	    sid_resource_create_signal_event_source(res, NULL, SIGHUP, NULL, 0, "sighup", NULL) < 0 ||
+	    sid_resource_create_signal_event_source(res, NULL, SIGCHLD, NULL, 0, "sigchld", NULL) < 0) {
```
prajnoha, line 54: ...so here we should to use a handler for `SIGTERM` with the `sid_resource_exit_event_loop` call.

prajnoha (commented):
OK for the `SIGPIPE` handler in `usid`. Just curious - do you know why `udevd` was sending `SIGPIPE` to `usid`?

src/resource/worker-control.c
```diff
@@ -693,23 +693,8 @@ sid_resource_t *worker_control_get_new_worker(sid_resource_t *worker_control_res
...
-			 *        See also https://github.com/sid-project/sid/issues/33.
-			 *
-			 *        For now, we just comment out the sid_resource_destroy which
-			 *        destroys the unneeded and inherited "sid" resource from parent
-			 *        daemon process. But we should fix this correctly and we should
-			 *        know why this is causing problems!
-			 */
-			/*(void) sid_resource_destroy(sid_resource_search(worker_control_res, SID_RESOURCE_SEARCH_TOP, NULL, NULL));*/
+			/* destroy the unneeded and inherited "sid" resource from parent */
+			(void) sid_resource_destroy(sid_resource_search(worker_control_res, SID_RESOURCE_SEARCH_TOP, NULL, NULL));
```
prajnoha, line 685: Yes! Hooray! :)

src/resource/worker-control.c
```diff
@@ -643,17 +642,6 @@ sid_resource_t *worker_control_get_new_worker(sid_resource_t *worker_control_res
...
-	if (sigfillset(&new_sigmask) < 0) {
-		log_sys_error(ID(worker_control_res), "sigfillset", "");
-		goto out;
-	}
-
-	if (sigprocmask(SIG_SETMASK, &new_sigmask, &original_sigmask) < 0) {
-		log_sys_error(ID(worker_control_res), "sigprocmask", "blocking signals before fork");
-		goto out;
-	}
-	signals_blocked = 1;
```
prajnoha, line 655: But here we were blocking all possible signals. The other parts of the code block only certain subset of the whole signal set.

The `worker-control` should encapsulate everything that is needed for the workers to get created and managed without any further dependencies and intervention from outside, so the outside code simply doesn't need to bother. IOW, the code inside `worker-control` shouldn't probably assume that the "outside" has done everything that `worker-control` needs for its proper functionality.

src/resource/sid.c
trgill: yes, ok - that would be better

src/resource/resource.c
trgill: Ok - will do

src/resource/worker-control.c
trgill: I see.  I'll drop these changes.

src/resource/sid.c
trgill: The function signature for sid_resource_exit_event_loop():

int sid_resource_exit_event_loop(sid_resource_t *res)

The signal handler is:

int _sid_signal_handler(sid_resource_event_source_t *es, const struct signalfd_siginfo *si, void *data)

I updated sid_resource_exit_event_loop() to take sid_resource_event_source_t * as the first parameter.  I'm not sure that was what you were looking to do.  Would it be better to do the parameter translation in the new _sid_signal_handler()?

src/resource/resource.c
trgill: moved to resource/sid.c - I think that is what you were asking

src/resource/sid.c
prajnoha: Or we can simply pass the resource as the custom arg to the signal handler function like here https://github.com/sid-project/sid/blob/f64443e69f3a824ea1433debcf9ad98bebaeed3b/src/resource/worker-control.c#L1098

For the `sid` resource (`src/resource/sid.c`), that could look something like this:

```
static int _on_sid_signal_event(sid_resource_event_source_t *es, const struct signalfd_siginfo *si, void *arg)
{
  sid_resource_t *res = arg;
  sid_resoruce_exit_event_loop(res);
}

....

static int _init_sid(sid_resource_t *res, const void *kickstart_data, void **data)
{
...
sid_resource_create_signal_event_source(res, NULL, SIGTERM, NULL, 0, "sigterm", res);
...
}
```

I've already used it this way in

trgill: Thanks.  That is much better.

PR #55, opened 2021-01-25
global: Use libelogind when libsystemd is not available

### LINKED ISSUE #17: Consider making dependency on systemd optional

tbzatek:
Elogind ships with the libelogind library that is built upon
libsystemd sources. Symbol names, header file naming and most
of its original functionality is preserved.

https://github.com/elogind/elogind

Resolves #17

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -86,7 +86,9 @@ AC_TYPE_UINT64_T
 AC_FUNC_ERROR_AT_LINE
 AC_FUNC_MALLOC
 
-PKG_CHECK_MODULES(SYSTEMD, libsystemd >= 221) # event loop supported in libsystemd version >= 221
+# event loop supported in libsystemd version >= 221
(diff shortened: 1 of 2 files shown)


tbzatek:
_(reposting after master branch name change, the old one: #32)_

PR #59, opened 2021-02-05
add tests for hash table functions

trgill:


src/base/hash.c
```diff
@@ -26,7 +26,7 @@
 
 struct hash_node {
 	struct hash_node *next;
-	void *data;
+	const void *data;
```
prajnoha, line 29: This looks cleaner, yes - that we instruct that the data stored inside the hash node is not actually changed subsequently so we can be sure that what was stored is then retrieved in exact same form.

However, this will probably require further changes in the code that uses this hash interface. Mainly, there's the `hash_update_binary` fn that we'd need to change here too I think - or the calling code to be more precise. The `hash_update_binary` is special in a way that before it does the actual update, it calls given callback to decide whether the record should be really updated or the old record should be kept.

 For example https://github.com/sid-project/sid/blob/846ddf290a0aea4ef3bdcc69f2a8f640b8f6d99c/src/resource/kv-store.c#L288-L295

Which is the case where the hash node's `data` stores `struct kv_store_value` instance. Based on the flags, the `struct kv_store_value` is capable of storing a raw pointer.

When we change the pointer by means of `hash_update_binary` call without changing any flags, we don't need to recreate the whole `struct kv_store_value`, we just need to change the pointer value inside it...

If the `data` in the hash node was strictly `const`, then we'd lose possibility to do this quick operation/update and we'd always need to discard old and create new values (the 'data') on value updates.

So it's something to consider whether making the data strictly 'const' is worth it... What we gain and what we lose by introducing more strictness on hash node's 'data' immutability.

prajnoha (commented):
Currently, there are 3 callers of this hash code that would require patching accordingly:

```
kv-store.c:378:14: warning: assigning to 'struct kv_store_value *' from 'const void *' discards qualifiers [-Wincompatible-pointer-types-discards-qualifiers]
        if (!(found = hash_lookup(kv_store->ht, key)))
                    ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
kv-store.c:401:14: warning: assigning to 'struct kv_store_value *' from 'const void *' discards qualifiers [-Wincompatible-pointer-types-discards-qualifiers]
        if (!(found = hash_lookup(kv_store->ht, key)))
                    ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
kv-store.c:434:14: warning: assigning to 'struct kv_store_value *' from 'const void *' discards qualifiers [-Wincompatible-pointer-types-discards-qualifiers]
        if (!(value = iter->current ? hash_get_data(iter->store->ht, iter->current) : NULL))
                    ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

But I'm afraid that these are just starting points. Once we start adding `const` here, we end up unwinding other chains of calls where we'd need to add the `const`. In at least one case I know about (the `hash_update_binary` call as I already mentioned in the other comment), we'd even need to change the logic a bit.

src/base/hash.c
trgill: @prajnoha I see - thanks.  Sorry I missed those warnings I introduced.

I'll drop the commit that adds the const.

PR #88, opened 2021-03-03
Add --coverage option to compile and link lines

trgill:
Allow a user to specify --enable-codecov on the configure command line
that translates to --coverage being passed to the compile and link
lines.  After sid is compiled, .gcda and .gcno files are generated.
HTML can then be generated via:

gcovr . --html-details > coverage.html

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -148,3 +148,9 @@ AC_SUBST([udevrulesdir])
 
 AC_OUTPUT
 AM_CHECK_CMOCKA
+
+AC_ARG_ENABLE([codecov],
+  [AS_HELP_STRING([--enable-codecov], [enable code covergae testing ])],,
+  [enable_codecov=yes])
+
+AM_CONDITIONAL([ENABLE_CODECOV], [test x$enable_codecov = xyes])


trgill:
@prajnoha this isn't ready to merge yet.  I need to make some updates to the Jenkins job.

Is there a better way to pass the --coverage flag down the the compile lines without updating every Makefile.am?  This was the best that I could get working.

prajnoha:
> @prajnoha this isn't ready to merge yet. I need to make some updates to the Jenkins job.
> 
> Is there a better way to pass the --coverage flag down the the compile lines without updating every Makefile.am? This was the best that I could get working.

I think something like this directly in `configure.ac` should do the job (...set conditionally when `configure --enable-codecov` is used):

```
CFLAGS="$CFLAGS --coverage"
```

So simply adding the --coverage to existing global `$CFLAGS` variable that should be honored globally then.

prajnoha:
...or defining extra variable in configure with flags that would be used in all Makefiles besides the usual `CFLAGS`. Then any further new configuration options adding more compilation options would just add to this extra variable, not fiddling with the `CFLAGS` which are dedicated for users to define, not the (configuration) script. We'd need to add this new variable to every makefile now, but then whenever adding more compilation options in the future, we'll have it in place already.

There are some hints about this here: https://www.gnu.org/software/automake/manual/html_node/Flag-Variables-Ordering.html.

prajnoha:
@trgill Todd, just wanted to ask about this one - it got closed, but we're not compiling with coverage support yet - were there any problems that you hit? Or we will revisit this later?

trgill:
@prajnoha I enabled code coverage in the Jenkins job rather than changing our build process.  For each PR on F33 we now:

sudo CFLAGS="--coverage -g -O0" LDFLAGS="--coverage" make clean;make

We then run most of the tests and it creates a coverage report with each job.  The report gets archived with the .log file of the Jenkins job.

I did run into some problems.

1. sidctl dump hangs on Centos Stream (opened issue #90) - this happens even without the --coverage flag
2. test_usid hangs when we've got --coverage enabled on F33 (opened issue #91)

We don't want our Jenkins tasks hanging.  It ties up nodes and requires manual intervention.  So, I worked around the two problems in our Jenkins job - but I will investigate and fix so we can get the entire test suite running on all (rawhid, F33, and Centos Stream).  After that, we can work to improve coverage.

Earlier this week there were some problems with DNS on the Jenkins systems, so I started working on the "tree" feature.  I think I'll wrap that up and then come back and fix or at least understand what is wrong with the issues above.

That sound ok to you?

trgill:
@prajnoha a couple other things to consider...

We could easily push our coverage HTML and logs to:

https://fedorapeople.org/groups/storage_apis/statuses/

That would make the Jenkins job results public for any non RH folks that get involved.

Maybe it would make sense to add instructions for coverage reports somewhere?

prajnoha:
Sure, thanks for looking into all of this! I just saw this one closed so wasn't sure :)  As for the instructions, we could possibly add a 'docs' directory directly inside the repo and add any hints and things we need documented and available with the source...

PR #89, opened 2021-03-09
add sidctl "tree" command to display resource tree in JSON

trgill:


src/resource/ubridge.c
```diff
@@ -516,6 +516,36 @@ static const char *_get_iov_str(struct buffer *buf, bool unset, struct iovec *io
...
+	int                     r = 0;
+	uint32_t                len;
+	void *                  data;
+
+	if (!(iter = sid_resource_iter_create(res))) {
+		log_error(ID(res), INTERNAL_ERROR "%s: failed to create record iterator", __func__);
+		return -ENOMEM;
+	}
+
+	sid_resource_dump_all_in_dot(sid_resource_search(res, SID_RESOURCE_SEARCH_TOP, NULL, NULL));
```
trgill, line 536: @prajnoha the data that is traversed in sid_resource_dump_all_in_dot() is what I should pack up to send to sidctl, right?

Currently it looks like this:

`[27456:27464]:resource.c:1038:sid_resource_dump_all_in_dot      <DOT> digraph resources {
[27456:27464]:resource.c:1039:sid_resource_dump_all_in_dot      <DOT> "worker/a29cafe8-ea1f-4850-8ccf-4877927e1a10";
[27456:27464]:resource.c:1007:_dump_children_recursively_in_dot <DOT> "aggregate/modules";
[27456:27464]:resource.c:1024:_dump_children_recursively_in_dot <DOT> "worker/a29cafe8-ea1f-4850-8ccf-4877927e1a10" -> "aggregate/modules" [dir=both];
[27456:27464]:resource.c:1007:_dump_children_recursively_in_dot <DOT> "module-registry/block";
[27456:27464]:resource.c:1024:_dump_children_recursively_in_dot <DOT> "aggregate/modules" -> "module-registry/block" [dir=both] [color=red];
[27456:27464]:resource.c:1007:_dump_children_recursively_in_dot <DOT> "module/blkid";
[27456:27464]:resource.c:1024:_dump_children_recursively_in_dot <DOT> "module-registry/block" -> "module/blkid" [dir=both] [color=red];
[27456:27464]:resource.c:1007:_dump_children_recursively_in_dot <DOT> "module/dummy_block";
[27456:27464]:resource.c:1024:_dump_children_recursively_in_dot <DOT> "module-registry/block" -> "module/dummy_block" [dir=both] [color=red];
[27456:27464]:resource.c:1007:_dump_children_recursively_in_dot <DOT> "module-registry/type";
[27456:27464]:resource.c:1024:_dump_children_recursively_in_dot <DOT> "aggregate/modules" -> "module-registry/type" [dir=both] [color=red];
[27456:27464]:resource.c:1007:_dump_children_recursively_in_dot <DOT> "module/dummy_type";
[27456:27464]:resource.c:1024:_dump_children_recursively_in_dot <DOT> "module-registry/type" -> "module/dummy_type" [dir=both] [color=red];
[27456:27464]:resource.c:1007:_dump_children_recursively_in_dot <DOT> "kv-store/main";
[27456:27464]:resource.c:1024:_dump_children_recursively_in_dot <DOT> "worker/a29cafe8-ea1f-4850-8ccf-4877927e1a10" -> "kv-store/main"  [dir=forward];
[27456:27464]:resource.c:1007:_dump_children_recursively_in_dot <DOT> "connection";
[27456:27464]:resource.c:1024:_dump_children_recursively_in_dot <DOT> "worker/a29cafe8-ea1f-4850-8ccf-4877927e1a10" -> "connection" [dir=both];
[27456:27464]:resource.c:1007:_dump_children_recursively_in_dot <DOT> "command/27464/tree";
[27456:27464]:resource.c:1024:_dump_children_recursively_in_dot <DOT> "connection" -> "command/27464/tree" [dir=both];
[27456:27464]:resource.c:1041:sid_resource_dump_all_in_dot      <DOT> }`

prajnoha: Yes, indeed, that's the core information we have about resources:
  - resource type
  - id
  - parent/children dependency
  - flags (for now it's various `SID_RESOURCE_RESTRICT_...` flags and `SID_RESOURCE_DISALLOW_ISOLATION` flag)

We could possibly add more insights from the core `struct sid_resource` like:
  - indicator of event loop existence for the resource
  - number of event sources attached
  - PID where the resource got created

(...that's what can be read from looking at `struct sid_resource` directly)

And then it would be great if we had a possibility (or at least count with this for the future) for each resource to send its own  extra information in some predefined format (I think a simple key-value should suffice, but also supporting arrays). Each resource could have a hook that we can call to do the per-resource custom dump (the declaration of new `dump` hook would go into `typedef struct sid_resource_type` besides `init` and `destroy` functions we already have there). We don't need to do this in one go, of course, just counting with this for the future that there might be more fields to add in the dump...

All of this this functionality should be mainly for us to have a better base for debugging and seeing state at certain point in time - like an introspection on demand without a need to run gdb or parsing through tons of logs.

prajnoha: (Of course, the dot is just an output format to view this graphically - but it's one of many possible. If we had some standard format for the resource dump - like json (we could reuie existing functions to print the json we have in `usid` now), then we could easily translate to specific output formats like dot or whatever else we needed...)

prajnoha: (...going even deeper with this later, we can even consider making the database dump just a custom `kv-store` resource dump that's simply added as a "subgraph" to the whole resource-tree dump)

trgill: @prajnoha something like this look ok to you?

I have some questions for you about how it should be implemented.  I'm currently generating the JSON in sid - wasn't sure what you'd think of that.... I'll push the code ASAP.

```
# sidctl tree --format json
{
    "ID": "worker/ca13a375-a97d-4dbf-9701-4b56ce9f5140",
    "type": "worker",
    "direction": "[dir=both]",
    "event-sources": [
    {
        "name": "main"
    },
    {
        "name": "worker_signal_handler"
    }
    ],
    "pid-created": 368550,
    "flags": 0,
    "prio": 0,
    "color": "",
    "children": [
        {
            "ID": "aggregate/modules",
            "type": "aggregate",
            "direction": "[dir=both]",
            "pid-created": 368543,
            "flags": 0,
            "prio": 0,
            "color": "",
            "children": [
                {
                    "ID": "module-registry/block",
                    "type": "module-registry",
                    "direction": "[dir=both]",
                    "pid-created": 368543,
                    "flags": 4,
                    "prio": 0,
                    "color": " [color=red]",
                    "children": [
                        {
                            "ID": "module/blkid",
                            "type": "module",
                            "direction": "[dir=both]",
                            "pid-created": 368543,
                            "flags": 4,
                            "prio": 0,
                            "color": " [color=red]"
                        },
                        {
                            "ID": "module/dummy_block",
                            "type": "module",
                            "direction": "[dir=both]",
                            "pid-created": 368543,
                            "flags": 4,
                            "prio": 1,
                            "color": " [color=red]"
                        }
                    ]
                },
                {
                    "ID": "module-registry/type",
                    "type": "module-registry",
                    "direction": "[dir=both]",
                    "pid-created": 368543,
                    "flags": 4,
                    "prio": 0,
                    "color": " [color=red]",
                    "children": [
                        {
                            "ID": "module/dummy_type",
                            "type": "module",
                            "direction": "[dir=both]",
                            "pid-created": 368543,
                            "flags": 4,
                            "prio": 1,
                            "color": " [color=red]"
                        }
                    ]
                }
            ]
        },
        {
            "ID": "kv-store/main",
            "type": "kv-store",
            "direction": " [dir=forward]",
            "pid-created": 368543,
            "flags": 1,
            "prio": 0,
            "color": ""
        },
        {
            "ID": "connection",
            "type": "connection",
            "direction": "[dir=both]",
            "event-sources": [
            {
                "name": "client connection"
            }
            ],
            "pid-created": 368550,
            "flags": 0,
            "prio": 0,
            "color": "",
            "children": [
                {
                    "ID": "command/368550/tree",
                    "type": "command",
                    "direction": "[dir=both]",
                    "event-sources": [
                    {
                        "name": "command handler"
                    }
                    ],
                    "pid-created": 368550,
                    "flags": 0,
                    "prio": 0,
                    "color": ""
                }
            ]
        }
    ]
}
```

trgill: test

prajnoha: Sending the JSON directly from `sid` to `sidctl` is a possibility. We just need to take into account two things:

  - JSON is just one of possible representations and in the future, we may support more formats. We'd either need to be able to send those different formats OR take the JSON as primary and do the translation on `sidctl` side to any other format. But I think that if we can format in JSON on `sid` side, then we can do any other format there as well. So I think then we need some bits of protocol parts that handle this - "please send me the result in format XYZ". This will probably make it easier for us to support other clients than `sidctl` itself.

  - We'll be sending a bit more data over the socket with JSON as that is text format (compared to binary). But on the other hand, I think we'll save some cycles elsewhere with this - we'll format this just once and then send over to `sidctl` for display (...compared to formatting to binary format on `sid` side, then sending it over to `sidctl`, then formatting it again to target format there).

So I think formatting on `sid` side is a good idea, just needs the extra bits in the protocol so we can select the format in which `sid` will send us the result.

prajnoha: Looking at the example output - maybe we don't need to send/print the "direction" as that is already encoded in "flags" field? Also, for those flags, it would probably be better if we printed the flag names (`NO_FLAGS`/`RESTRICT_WALK_UP`/`RESTRICT_WALK_DOWN`/`DISALLOW_ISOLATION`). This could prevent confusion if we needed to reorder the flags later...

trgill: @prajnoha Ok - I removed the direction and color fields from the output.

I haven't added the bit of the protocol parts.  I was thinking we might want to consider using a JSON API.  Maybe make the socket that SID listens the API for SID?  Does that seem like a reasonable idea to you?

src/resource/ubridge.c
```diff
@@ -3434,10 +3434,6 @@ static int _cmd_handler(sid_resource_event_source_t *es, void *data)
 		return -1;
 	}
 
-	if ((r = _export_kv_store(cmd_res)) < 0) {
```
trgill, line 3437: @prajnoha am I missing something here?  It looks like we append the kv store even when responding to commands other than "dump"?

src/resource/ubridge.c
prajnoha: Sure, I'm counting with a connection point/socket for the API (now I mean the other API for all the other commands and queries, not the API used to construct SID `ucmd` modules - these are executed inside SID daemon directly and they communicate through the DB itself).

As for JSON API - I think it's going to be fine for almost 99% of our needs. The only area where I'm a bit wary is sending more data, like the db dump which can grow easily with the number of devices growing on the system. But on the other hand - it's just a global whole db dump used for debug, not the usual use case. The usual use cases are not, and I believe won't be, formatting and transferring so much data (instead, working on subset of the information recorded in db, for example: "send me all devices affected by change on device XYZ in the stack").

The other thing is also a possibility to encode binary data (e.g. the database can contain binary data, we're not bound only to strings... though again, they're strings most of the time.). I'm not sure right now how JSON format handles such cases.

In summary, I'm definitely fine with JSON API and surely sounds reasonable - it will make the connection with other services much easier. We just need to be aware of those two things mentioned above...

src/resource/ubridge.c
prajnoha: Ah yes, this is something I've also noticed recently - it's actually a bug!

The `_export_kv_store` is supposed to export the KV store snapshot diffs from SID worker process and export it to main process for syncing with main DB. Of course, this (currently) makes sense only for the `usid scan` command and we shouldn't be calling `_export_kv_store` for any other cases, including any of `sidctl` commands, because they do not change the database in any way, just peeking at it.

I have a patch for this in my queue, will commit that...

prajnoha: ...well, for records marked as belonging to `KV_NS_UDEV` namespace, the `_export_kv_store` exports to udev db actually, not the SID db. But all the other namespaces (`KV_NS_DEVICE`, `KV_NS_MODULE`, `KV_NS_GLOBAL`) do sync with main SID db. So from this perspective, the SID db is extension to udev db in a way.

trgill: Ok - thanks for confirming.  It causes problems with the JSON response.  When sidctl gets the response, the size includes the data that is added via _export_kv_store().

src/resource/ubridge.c
trgill: It looks like we could encode binary data via base64.  base64 would increase the data size by ~ 33%.

src/resource/ubridge.c
prajnoha: This should be patched within PR #96. Please try with that patch...

trgill: Ok - will test now.

trgill: Hold on.... 

I made a mistake testing and ran an old version (or something).  It is working.

src/tools/sidctl/sidctl.c
```diff
@@ -329,36 +159,36 @@ static int _usid_cmd_stats(struct args *args, output_format_t format)
...
-	print_uint_field(KEY_SIDCTL_MINOR, SID_VERSION_MINOR, format, true, 1);
-	print_uint_field(KEY_SIDCTL_RELEASE, SID_VERSION_RELEASE, format, r == 0, 1);
+	print_uint_field(KEY_SIDCTL_PROTOCOL, USID_PROTOCOL, format, outbuf, true, 1);
+	print_uint_field(KEY_SIDCTL_MAJOR, SID_VERSION_MAJOR, format, outbuf, true, 1);
+	print_uint_field(KEY_SIDCTL_MINOR, SID_VERSION_MINOR, format, outbuf, true, 1);
+	print_uint_field(KEY_SIDCTL_RELEASE, SID_VERSION_RELEASE, format, outbuf, r == 0, 1);
 
 	if (r == 0) {
-		buffer_get_data(buf, (const void **) &hdr, &size);
+		buffer_get_data(outbuf, (const void **) &hdr, &size);
```
prajnoha, line 200: When I tried `sidctl version`, I got incorrect output for SID version - it's just this line that has typo, I think. It should be `readbuf` instead of `outbuf` here.

src/tools/sidctl/sidctl.c
```diff
@@ -48,6 +48,27 @@ struct args {
...
+	int                       r;
+
+	if ((r = usid_req(LOG_PREFIX, USID_CMD_TREE, 0, NULL, NULL, &readbuf)) == 0) {
+		buffer_get_data(readbuf, (const void **) &msg, &size);
+		if (size < USID_MSG_HEADER_SIZE || msg->status & COMMAND_STATUS_FAILURE) {
+			buffer_destroy(readbuf);
+			return -1;
+		}
+		size -= USID_MSG_HEADER_SIZE;
+		buffer_add(outbuf, msg->data, size, &r);
```
prajnoha: Hmm, we're reading almost the whole complete and formatted message here, just skipping the header. Then we're copying all the rest to the output buffer. It works fine, but would be nicer if we could avoid the extra copy.

Maybe we could define the output buffer as `BUFFER_TYPE_VECTOR` in which case we could add the pointer to the `msg->data` + `size` directly into the vector buffer as element. But right now, formatting is done on sidctl side for other command now and they use the output buffer for formatting so there the `BUFFER_TYPE_VECTOR` is not ideal. But this is caused by the fact that `tree` command is formatted on SID side while the rest is formatted on `sidctl` side at the moment.

So maybe we can keep this as it is for now (copying the `msg->data` to `outbuf` which is `BUFFER_TYPE_LINEAR`), later we can optimize and/or move the rest of commands to SID-side formatting.

prajnoha (commented):
OK, the patchset looks good. Just these trivia:

```
ubridge.c:1716:43: warning: unused variable 'data' [-Wunused-variable]
        char *               resource_tree_data, data;
                                                 ^
ubridge.c:1717:29: warning: unused variable 'buf_size' [-Wunused-variable]
        size_t               size, buf_size;
                                   ^
```

+ the comments in the code.

As for binary data - yes, we could use base64 which is pretty standard and it's easy to decode. And also we could simply print those values - we'd need some representation of those binary data anyway for printing. I don't expect we'll ever store huge amounts of binary data, they're usually short chunks of direct binary structures so the extra cost of ~33% is OK I think.

src/tools/sidctl/sidctl.c
trgill: Ok.  I'll open an issue to fix.

src/tools/sidctl/sidctl.c
trgill: Thanks.  Fixed.

src/tools/sidctl/sidctl.c
trgill: Fixed the warnings.  Sorry for that mistake.

prajnoha:
OK, merged, thanks! There's just one thing - `SID_SESSION_ID=...` added at the very end of output, but I think I know why it's there - I need to exclude the `sidctl tree` cmd from the ones that generate/can use the session id - I'll patch that, should be one liner.

trgill:
:+1:

PR #92, opened 2021-03-17
resource: Don't create a signal event for SIGCHLD

bmarzins:
Signal event handling through sd_event_add_signal() has special code to
deal with the fact that sd_event_add_child() may also be watching for
the SIGCHLD signal. If SIGCHLD is watched by the new SID signal event
code, it can get stuck in _sd_signal_event_handler() when a child exits.

Don't watch for SIGCHLD, to avoid interfering with sd_event_add_child().
Also, open the signalfd in non-blocking mode, as sd_event_add_signal()
does, to avoid the possibility of getting stuck in
_sd_signal_event_handler().


prajnoha:
Good catch, thanks for the patch!

PR #96, opened 2021-03-26
resource: ubridge: define and use flags for _cmd_regs representing ucmd commands

prajnoha:
Define CMD_KV_IMPORT_UDEV, CMD_KV_EXPORT_UDEV, CMD_KV_EXPORT_SID,
CMD_SESSION_ID to control whether certain records are imported/exported
and whether the command can make use and export session ID (support for
sessions is yet to be implemented). This is mainly to avoid trying to
import and export records for commands which are not supposed to do so.

### DIFF
--- a/src/resource/ubridge.c
+++ b/src/resource/ubridge.c
@@ -290,7 +290,15 @@ struct cross_bitmap_calc_arg {
 };
 
 /*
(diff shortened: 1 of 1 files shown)


prajnoha:
Included in PR #89.

PR #99, opened 2021-04-07
sid: add prctl(PR_SET_PDEATHSIG, SIGUSR1) to SID startup

trgill:
If the parent SID process is terminated, send a SIGINT to any active worker processes so they terminate.

### DIFF
--- a/src/resource/worker-control.c
+++ b/src/resource/worker-control.c
@@ -676,6 +676,7 @@ sid_resource_t *worker_control_get_new_worker(sid_resource_t *worker_control_res
 	struct worker_channel * worker_proxy_channels = NULL, *worker_channels = NULL;
 	struct worker_kickstart kickstart;
 	sigset_t                original_sigmask, new_sigmask;
+	pid_t                   original_pid, curr_ppid;
 	sid_resource_t *        res             = NULL;
 	int                     signals_blocked = 0;
 	pid_t                   pid             = -1;
@@ -699,6 +700,7 @@ sid_resource_t *worker_control_get_new_worker(sid_resource_t *worker_control_res
 		goto out;
 	}
 	signals_blocked = 1;
+	original_pid    = getpid();
 
 	if ((pid = fork()) < 0) {
 		log_sys_error(ID(worker_control_res), "fork", "");
@@ -713,6 +715,14 @@ sid_resource_t *worker_control_get_new_worker(sid_resource_t *worker_control_res
 		/* request a SIGUSR1 signal if parent dies */
 		if (prctl(PR_SET_PDEATHSIG, SIGUSR1) < 0)
 			log_sys_error(ID(worker_control_res), "prctl() failed", "");
+
+		/* Check to make sure the parent didn't die right after the fork() */
+		curr_ppid = getppid();
+		if (curr_ppid != original_pid) {
+			log_debug(ID(worker_control_res), "parent died before prctl() call completed - exiting SID worker process");
+			raise(SIGTERM);
+			exit(EXIT_FAILURE);
+		}
 
 		_destroy_channels(worker_proxy_channels, worker_control->channel_spec_count);
 		worker_proxy_channels = NULL;
--- a/src/resource/worker-control.c
+++ b/src/resource/worker-control.c
@@ -26,6 +26,7 @@
 #include "log/log.h"
 #include "resource/resource.h"
 
+#include <sys/prctl.h>
 #include <unistd.h>
 
 #define WORKER_CONTROL_NAME "worker-control"
@@ -98,6 +99,7 @@ struct worker {
 	struct worker_channel_spec *channel_specs;
 	struct worker_channel *     channels; /* NULL-terminated array of worker --> worker_proxy channels */
 	unsigned                    channel_count;
+	unsigned                    parent_exited;
 };
 
 static void _change_worker_proxy_state(sid_resource_t *worker_proxy_res, worker_state_t state)
@@ -708,6 +710,10 @@ sid_resource_t *worker_control_get_new_worker(sid_resource_t *worker_control_res
 		 *  WORKER HERE
 		 */
 
+		/* request a SIGUSR1 signal if parent dies */
+		if (prctl(PR_SET_PDEATHSIG, SIGUSR1) < 0)
+			log_sys_error(ID(worker_control_res), "prctl() failed", "");
+
 		_destroy_channels(worker_proxy_channels, worker_control->channel_spec_count);
 		worker_proxy_channels = NULL;
 
@@ -841,7 +847,6 @@ out:
 	/* run event loop in worker's top-level resource */
 	if (r == 0)
 		r = sid_resource_run_event_loop(res);
-
 	if (res)
 		(void) sid_resource_destroy(res);
 
@@ -895,6 +900,7 @@ const char *worker_control_get_worker_id(sid_resource_t *res)
 /* FIXME: Consider making this a part of event loop. */
 static int _chan_buf_send(const struct worker_channel *chan, worker_channel_cmd_t cmd, struct worker_data_spec *data_spec)
 {
+	struct worker *      worker;
 	static unsigned char byte     = 0xFF;
 	int                  has_data = data_spec && data_spec->data && data_spec->data_size;
 	ssize_t              n;
@@ -1027,7 +1033,10 @@ int worker_control_worker_yield(sid_resource_t *res)
 	for (i = 0; i < worker->channel_count; i++) {
 		chan = &worker->channels[i];
 		if (chan->spec->wire.type == WORKER_WIRE_PIPE_TO_PROXY || chan->spec->wire.type == WORKER_WIRE_SOCKET)
-			return _chan_buf_send(chan, WORKER_CHANNEL_CMD_YIELD, NULL);
+			if (worker->parent_exited == 0)
+				return _chan_buf_send(chan, WORKER_CHANNEL_CMD_YIELD, NULL);
+			else
+				raise(SIGTERM);
 	}
 
 	return -ENOTCONN;
@@ -1083,9 +1092,23 @@ static int _on_worker_proxy_idle_timeout_event(sid_resource_event_source_t *es, 
 static int _on_worker_signal_event(sid_resource_event_source_t *es, const struct signalfd_siginfo *si, void *arg)
 {
 	sid_resource_t *res = arg;
+	struct worker * worker;
 
 	log_debug(ID(res), "Received signal %d from %d.", si->ssi_signo, si->ssi_pid);
-	sid_resource_exit_event_loop(res);
+
+	switch (si->ssi_signo) {
+		case SIGTERM:
+		case SIGINT:
+			sid_resource_exit_event_loop(res);
+			break;
+		case SIGUSR1:
+			worker = sid_resource_get_data(res);
+			if (worker != NULL)
+				worker->parent_exited = 1;
+			break;
+		default:
+			break;
+	};
 
 	return 0;
 }
@@ -1147,6 +1170,7 @@ static int _init_worker(sid_resource_t *worker_res, const void *kickstart_data, 
 	sigemptyset(&mask);
 	sigaddset(&mask, SIGTERM);
 	sigaddset(&mask, SIGINT);
+	sigaddset(&mask, SIGUSR1);
 
 	if (!(worker = mem_zalloc(sizeof(*worker)))) {
 		log_error(ID(worker_res), "Failed to allocate new worker structure.");
@@ -1156,6 +1180,7 @@ static int _init_worker(sid_resource_t *worker_res, const void *kickstart_data, 
 	worker->channel_specs = kickstart->channel_specs;
 	worker->channels      = kickstart->channels;
 	worker->channel_count = kickstart->channel_count;
+	worker->parent_exited = 0;
 
 	if (sid_resource_create_signal_event_source(worker_res,
 	                                            NULL,
--- a/src/resource/worker-control.c
+++ b/src/resource/worker-control.c
@@ -1163,7 +1163,7 @@ static int _init_worker(sid_resource_t *worker_res, const void *kickstart_data, 
 	                                            _on_worker_signal_event,
 	                                            0,
 	                                            "worker_signal_handler",
-	                                            NULL) < 0) {
+	                                            worker_res) < 0) {
 		log_error(ID(worker_res), "Failed to create signal handlers.");
 		goto fail;
 	}
--- a/src/resource/resource.c
+++ b/src/resource/resource.c
@@ -484,7 +484,7 @@ static int _sd_signal_event_handler(sd_event_source *sd_es, int sfd, uint32_t re
 		return 1;
 	}
 
-	return ((sid_resource_signal_event_handler_t) es->handler)(es, &si, es->res);
+	return ((sid_resource_signal_event_handler_t) es->handler)(es, &si, es->data);
 }
 
 /* This should not watch the SIGCHLD signal if sd_event_add_child() is also used */
@@ -534,7 +534,7 @@ int sid_resource_create_signal_event_source(sid_resource_t *                    
 		goto fail;
 	}
 
-	if ((r = _create_event_source(res, name, sd_es, handler, NULL, es)) < 0)
+	if ((r = _create_event_source(res, name, sd_es, handler, data, es)) < 0)
 		goto fail;
 
 	if (prio && (r = sd_event_source_set_priority(sd_es, prio)) < 0)
--- a/src/resource/sid.c
+++ b/src/resource/sid.c
@@ -65,7 +65,7 @@ static int _init_sid(sid_resource_t *res, const void *kickstart_data, void **dat
 	sigaddset(&mask, SIGPIPE);
 	sigaddset(&mask, SIGHUP);
 
-	if (sid_resource_create_signal_event_source(res, NULL, mask, _on_sid_signal_event, 0, "signal_handler", NULL) < 0) {
+	if (sid_resource_create_signal_event_source(res, NULL, mask, _on_sid_signal_event, 0, "signal_handler", res) < 0) {
 		log_error(ID(res), "Failed to create signal handlers.");
 		goto fail;
 	}


prajnoha:
Hmm, if it's an operation in worker process that doesn't necessarily require main process to be present anymore, we could probably just finish the operation and return the result back to the client. Like that `sidctl dump` - after forking the worker process from main process, we already have everything needed for the `dump` operation.

The only issue here (with completing the current operation in worker process even if the main process is gone) is that worker process may still try to call `worker_control_worker_yield`. This would end up with some error because the other endpoint is already closed. This should probably be "no operation" because there's no main process/manager that would listen to that `yield` request and so it makes no sense anymore. The `yield` in case of cmd handling in worker process is called here:

https://github.com/sid-project/sid/blob/65e3baf8721bcd14188ca54a2203aeed084e74ad/src/resource/ubridge.c#L3431-L3444

Of course, if it's any other operation that is needed, like syncing the db snapshot with main process like the `usid scan` command does (the `_export_kv_store` fn), we should fail correctly if the main process is already gone:

https://github.com/sid-project/sid/blob/65e3baf8721bcd14188ca54a2203aeed084e74ad/src/resource/ubridge.c#L3476-L3489

Looking at that code right now, I think we shouldn't even be returning any results to the client (the `buffer_write_all` call there). For returning the results to the client, we should probably wait for confirmation that main process has really merged the results into its own db. But maybe that's going too deep...

prajnoha:
> The only issue here (with completing the current operation in worker process even if the main process is gone) is that worker process may still try to call `worker_control_worker_yield`. This would end up with some error because the other endpoint is already closed. This should probably be "no operation" because there's no main process/manager that would listen to that `yield` request and so it makes no sense anymore. 

...maybe there should be a hook directly in `worker-control` code detecting that the main process is gone and store that state somewhere in `worker` struct. Then if the worker calls something (like that `worker_control_worker_yield`), we could simply check and return `-ESOMETHING` to denote that the main process is not there anymore.

I suppose there's a tiny race - the worker could try contacting the main process before it receives a signal notifying that the main process is actually gone. But I think we can't avoid that...

trgill:
@prajnoha could you please explain where we are going with the worker design?

There are functions and comments that make me think re-use of workers is intended.  For example:

```
sid_resource_t *worker_control_get_idle_worker(sid_resource_t *worker_control_res);
sid_resource_t *worker_control_find_worker(sid_resource_t *worker_control_res, const char *id);
/* Yield current worker and make it available for others to use. */
int worker_control_worker_yield(sid_resource_t *res);
```

But, currently workers are not re-used, right?  Are we going to change that?

I'm basing this comment on running commands with a breakpoint set at: 

https://github.com/sid-project/sid/blob/4aa2bdd01fbc4c5144ad5aec3578430b82707600/src/resource/ubridge.c#L4180

It is hit for all the commands that I've run.

Also, are we thinking we are going to sync the database to disk or is the always going to be in memory?

trgill:
> > The only issue here (with completing the current operation in worker process even if the main process is gone) is that worker process may still try to call `worker_control_worker_yield`. This would end up with some error because the other endpoint is already closed. This should probably be "no operation" because there's no main process/manager that would listen to that `yield` request and so it makes no sense anymore.
> 
> ...maybe there should be a hook directly in `worker-control` code detecting that the main process is gone and store that state somewhere in `worker` struct. Then if the worker calls something (like that `worker_control_worker_yield`), we could simply check and return `-ESOMETHING` to denote that the main process is not there anymore.
> 
> I suppose there's a tiny race - the worker could try contacting the main process before it receives a signal notifying that the main process is actually gone. But I think we can't avoid that...

If we want to track the parent exit in the child, we shouldn't use SIGTERM (like I did), right?  We'd need something unique to indicate parent exit.

prajnoha:
Hi! Yes, right, I'd like to keep a worker pool and then reuse workers from it. But that will require more work on the database part so we can create the snapshots not only based on COW feature of `fork` (it might be even a different backend for the database in the future - reusing an existing "real" database), but on demand directly from the worker. The database should also be synced to disk in the future (for possible reuse over reboots etc.). 

Right now, we're not reusing workers and not syncing to disk.

The other function - `worker_control_find_worker` - is primarily for us to be able to connect to the same "session", meaning if we called several `usid` commands within one uevent processing (for example `usid scan` and `usid checkpoint`).

The `SIGTERM` - yeah, indeed, that might be ambiguous because that's also the usual signal we use to terminate the process. So we should use some other signal like SIGUSR1? Or any other that fits better and can't be mistaken with any "common signal".

trgill:
@prajnoha I'm thinking we probably shouldn't be re-entering the main loop here:

https://github.com/sid-project/sid/blob/4c46c37a301872a28ac19e905f260b8116ce66de/src/resource/worker-control.c#L842

But maybe it makes sense to sort out syncing to disk/reusing workers first?

I've tested these changes by setting breakpoints in the worker and killing the parent - it looks to be working properly.

prajnoha:
> @prajnoha I'm thinking we probably shouldn't be re-entering the main loop here:
> 
> https://github.com/sid-project/sid/blob/4c46c37a301872a28ac19e905f260b8116ce66de/src/resource/worker-control.c#L842
> 

This is the new instance of the even loop created for the worker (the original one inherited from the parent process should be destroyed... well, ideally... but in real, systemd lib refuses to cleanup everything if it detects the PID has changed - so we can't do complete cleanup of the old event loop in the worker. Anyway, we're not using the original event loop in worker anymore, only the new one.)

We're trying to destroy the inherited event loop within this callback:

https://github.com/sid-project/sid/blob/4c46c37a301872a28ac19e905f260b8116ce66de/src/resource/worker-control.c#L748-L749

Then within the "worker init" callback, it's exactly this place where we destroy the original inherited event loop from parent:

https://github.com/sid-project/sid/blob/4c46c37a301872a28ac19e905f260b8116ce66de/src/resource/ubridge.c#L4115

The new event loop is then executed only for the worker:

https://github.com/sid-project/sid/blob/4c46c37a301872a28ac19e905f260b8116ce66de/src/resource/worker-control.c#L837-L843

prajnoha:
Looking at the code, I wonder if we could somehow even cover the window between the `fork` and the `prctl` call. It could possibly happen that the parent dies right in this window which isn't covered...

prajnoha:
> Looking at the code, I wonder if we could somehow even cover the window between the `fork` and the `prctl` call. It could possibly happen that the parent dies right in this window which isn't covered...

(Just looked how systemd deals with this, I can see they do this extra check:
https://github.com/systemd/systemd/blob/527653f827b41f9cd10ea47375b4d1533885b8e3/src/basic/process-util.c#L1387-L1400)

trgill:
> > Looking at the code, I wonder if we could somehow even cover the window between the `fork` and the `prctl` call. It could possibly happen that the parent dies right in this window which isn't covered...
> 
> (Just looked how systemd deals with this, I can see they do this extra check:
> https://github.com/systemd/systemd/blob/527653f827b41f9cd10ea47375b4d1533885b8e3/src/basic/process-util.c#L1387-L1400)

Thanks.  Will update.

trgill:
> 
> The new event loop is then executed only for the worker:
> 
> https://github.com/sid-project/sid/blob/4c46c37a301872a28ac19e905f260b8116ce66de/src/resource/worker-control.c#L837-L843

I'm not sure I understand.

I think you already know... when we enter that call to sid_resource_run_event_loop() this is the call stack:

libsidresource.so.0!worker_control_get_new_worker(sid_resource_t * worker_control_res, struct worker_params * params) (/home/todd/workspace/sid/src/resource/worker-control.c:843)
libsidresource.so.0!_on_ubridge_interface_event(sid_resource_event_source_t * es, int fd, uint32_t revents, void * data) (/home/todd/workspace/sid/src/resource/ubridge.c:4148)
libsidresource.so.0!_sd_io_event_handler(sd_event_source * sd_es, int fd, uint32_t revents, void * data) (/home/todd/workspace/sid/src/resource/resource.c:434)
libsystemd.so.0!source_dispatch(sd_event_source * s) (/home/todd/workspace/systemd/src/libsystemd/sd-event/sd-event.c:3498)
libsystemd.so.0!sd_event_dispatch(sd_event * e) (/home/todd/workspace/systemd/src/libsystemd/sd-event/sd-event.c:3950)
libsystemd.so.0!sd_event_run(sd_event * e, uint64_t timeout) (/home/todd/workspace/systemd/src/libsystemd/sd-event/sd-event.c:4011)
libsystemd.so.0!sd_event_loop(sd_event * e) (/home/todd/workspace/systemd/src/libsystemd/sd-event/sd-event.c:4032)
libsidresource.so.0!sid_resource_run_event_loop(sid_resource_t * res) (/home/todd/workspace/sid/src/resource/resource.c:984)
main(int argc, char ** argv) (/home/todd/workspace/sid/src/daemon/sid.c:178)

Just not sure it is safe to push another event loop on top of a destroyed loop? I understand we could avoid unwinding that call stack by just exiting - but concerned there are potential issues with this approach when we implement process reuse.

prajnoha:
> I think you already know... when we enter that call to sid_resource_run_event_loop() this is the call stack:
> ```
> libsidresource.so.0!worker_control_get_new_worker(sid_resource_t * worker_control_res, struct worker_params * params) (/home/todd/workspace/sid/src/resource/worker-control.c:843)
> ...
> libsidresource.so.0!sid_resource_run_event_loop(sid_resource_t * res) (/home/todd/workspace/sid/src/resource/resource.c:984)
> main(int argc, char ** argv) (/home/todd/workspace/sid/src/daemon/sid.c:178)
> ```

Ah yes, sorry I misunderstood your comment. Sure, this issue, I recall...

Of course, relying on the fact that we're not unwinding the call stack in the worker is far from ideal and it's definitely a pain spot with possible side effects, yes.

This one's a bit tricky though - we can't exit event loop in the parent process (we'd lose events/requests) and we can't exit event loop in the worker (at least not with `sd-event` loop which checks whether the PID hasn't changed on almost all API calls). 

For a clean solution here, we'd either need to change underlying event loop to something else than `sd-event` that supports cleaning up after forking (maybe `libev`?) OR we'd need to create a minimal template process for creating new workers.

(There's also a 3rd option - create our own event loop implementation where we won't be restricted, but we'd need to gain all the know-how that was put into the existing implementations.)

With the "worker template process", the main process would pass the request to this template process and the template process would then fork itself in turn, creating a new worker (without any call stack remnants). The template process would use a very simple event loop to accept these requests ("simple" meaning just to accept those specific requests and easy to exit in forked process).

I think the approach with the "worker template process" is more appropriate as with that, we're not restricting the event loop library we could possibly use. Also, starting a new worker process from a clean template seems much more straightforward. Disadvantage is that it adds intermediary between the main process and worker process, but only during creation. When the workers are already created, the communication can be conducted directly between the main process and that worker process.

prajnoha:
> @prajnoha I'm thinking we probably shouldn't be re-entering the main loop here:
> 
> https://github.com/sid-project/sid/blob/4c46c37a301872a28ac19e905f260b8116ce66de/src/resource/worker-control.c#L842
> 
> But maybe it makes sense to sort out syncing to disk/reusing workers first?
> 

...so yes - worker reuse requires us to add more enhancements to the db code. We need to be able to take a db snapshot by different means than we do now (relying on the fork itself). Possibly by mmap-ing the db or reusing existing db that supports such snapshots.

> I've tested these changes by setting breakpoints in the worker and killing the parent - it looks to be working properly.

Yup, I think we can merge this.

src/resource/worker-control.c
```diff
@@ -35,6 +36,8 @@
 
 #define DEFAULT_WORKER_IDLE_TIMEOUT_USEC 5000000
 
+static int sid_parent_exited = 0;
```
prajnoha, line 39: Just realized this - we should add this flag somewhere in the `struct worker` directly, because we can have more than one instance of `worker-control` and then its instances would interfere with each other through this global variable.

src/resource/worker-control.c
```diff
@@ -1027,7 +1033,10 @@ int worker_control_worker_yield(sid_resource_t *res)
 	for (i = 0; i < worker->channel_count; i++) {
 		chan = &worker->channels[i];
 		if (chan->spec->wire.type == WORKER_WIRE_PIPE_TO_PROXY || chan->spec->wire.type == WORKER_WIRE_SOCKET)
-			return _chan_buf_send(chan, WORKER_CHANNEL_CMD_YIELD, NULL);
+			if (!sid_parent_exited)
+				return _chan_buf_send(chan, WORKER_CHANNEL_CMD_YIELD, NULL);
```
prajnoha, line 1047: There can be more channels between the `worker` and `worker_proxy` so we have one more `_chan_buf_send` call (in `worker_channel_send` besides `worker_control_worker_yield`).

To avoid the same check for "parent exited" in all these places, maybe we could just put that check inside `_chan_buf_send`. There we need to check whether we are trying to send through the channel from `worker` side, not `worker_proxy` side (we can use `worker_control_is_worker(chan->owner)`).

Maybe we could also send some unique error code that we can check for in the `_chan_buf_send` callers, like `-ENOLINK`. And when it happens, we can add a log message indicating this condition - so we always know that the `worker` tried to send something through the channel to the parent, but the link is gone...

prajnoha (changes requested):
> > I've tested these changes by setting breakpoints in the worker and killing the parent - it looks to be working properly.
> 
> Yup, I think we can merge this.

...realized a few more things that we should probably also address - please, see review comments.

src/resource/worker-control.c
trgill: @prajnoha do we have access to the struct worker in the _on_worker_signal_event()?

I'm stepping through the structures in the debugger and it isn't obvious to me what you asking.

prajnoha: I thought we are already passing the worker resource to the signal handler, because it's already used there:

https://github.com/sid-project/sid/blob/93563cd39dfd686028da274e3f972f1660b5dba0/src/resource/worker-control.c#L1083-L1088

But then I checked how this is actually set up and noticed there's a tiny bug, should be fixed with this: https://github.com/prajnoha/sid/commits/fix_resource_signal_handler_data

trgill:
I think the changes I pushed address your comments.

But there is still a problem with the code.  When the child process calls sid_resource_run_event_loop() the parent no longer exists, so the SIGINT doesn't get sent and the child waits.

I'm looking at the possibility of us implementing a main loop as you've suggested.

src/resource/worker-control.c
```diff
@@ -1082,12 +1095,31 @@ static int _on_worker_proxy_idle_timeout_event(sid_resource_event_source_t *es,
...
+    switch (si->ssi_signo) {
+        case SIGTERM:
+        case SIGINT:
+            sid_resource_exit_event_loop(res);
+            break;
+        case SIGUSR1:
+            if (sid_resource_match(res, &sid_resource_type_worker, NULL))
+                worker_res = res;
+            else if (!(worker_res = sid_resource_search(res, SID_RESOURCE_SEARCH_ANC, &sid_resource_type_worker, NULL)))
+                break;
```
prajnoha, line 1123: This check is probably not needed because we have `_on_worker_signal_event` always called with the `worker_res` passed in...

prajnoha:
> I think the changes I pushed address your comments.

Yup, looks good, thanks! Just one thing - we probably don't need the check in `_on_worker_signal_event` (see previous comment).

> But there is still a problem with the code. When the child process calls sid_resource_run_event_loop() the parent no longer exists, so the SIGINT doesn't get sent and the child waits.

But we have this sequence:

  - A1 blocking signals (in parent process)
  - A2 calling `fork` (in parent process)
  - A3 (some more parent code)

  - B1 calling `prctl` (in child process)
  - B2 checking whether parent hasn't died in between `fork` and `prctl` (in child process)
  - B3 creating worker resource, including registration of `_on_worker_signal_event` (in child process)
  - B4 unblocking signals (in child process)
  - B5 calling `sid_resource_run_event_loop` (in child process)

The blocked signals should be queued and then collected by the newly executed event loop, so I'd expect that to work... though looking at the sequence as I'm writing it here, I think we should call that `prctl` and the check for "parent not dead already" after B3 where we actually register the `_on_worker_signal_event` so we remove even the race window between B1 and B3.

> I'm looking at the possibility of us implementing a main loop as you've suggested.

You mean the template process? That would be great! BUT then we'd need to find a different way of creating the snapshot of the DB. We can't use the "fork will do a snapshot for us" idea anymore because the template would either have an outdated snapshot (from the very beginning of the SID execution when the template was created) or won't have any db attached at all. So we need to explore the possibilities here...

src/resource/worker-control.c
trgill: Ok - I see.  Thanks - removed.

trgill:
> > But there is still a problem with the code. When the child process calls sid_resource_run_event_loop() the parent no longer exists, so the SIGINT doesn't get sent and the child waits.
> 
> But we have this sequence:
> 
>     * A1 blocking signals (in parent process)
> 
>     * A2 calling `fork` (in parent process)
> 
>     * A3 (some more parent code)
> 
>     * B1 calling `prctl` (in child process)
> 
>     * B2 checking whether parent hasn't died in between `fork` and `prctl` (in child process)
> 
>     * B3 creating worker resource, including registration of `_on_worker_signal_event` (in child process)
> 
>     * B4 unblocking signals (in child process)
> 
>     * B5 calling `sid_resource_run_event_loop` (in child process)
> 
> 
> The blocked signals should be queued and then collected by the newly executed event loop, so I'd expect that to work...
> 
> ~Though looking at the sequence as I'm writing it here, I think we should call that `prctl` and the check for "parent not dead already" after B3 where we actually register the `_on_worker_signal_event` so we remove even the race window between B1 and B3.~ (well, probably not needed, depends on how the signals are actually picked up after being unblocked)
>

I think the child exiting is currently dependent on the parent reaching where the SIGTERM is sent in:

https://github.com/sid-project/sid/blob/93563cd39dfd686028da274e3f972f1660b5dba0/src/resource/worker-control.c#L319

That is reached via the following call stack.  It is dependent on the child sending the parent a response.  If the parent dies while the worker is running - it will never get the response to trigger:
```

libsidresource.so.0!_make_worker_exit(sid_resource_t * worker_proxy_res) (/home/todd/workspace/sid/src/resource/worker-control.c:326)
libsidresource.so.0!_on_worker_proxy_channel_event(sid_resource_event_source_t * es, int fd, uint32_t revents, void * data) (/home/todd/workspace/sid/src/resource/worker-control.c:364)
libsidresource.so.0!_sd_io_event_handler(sd_event_source * sd_es, int fd, uint32_t revents, void * data) (/home/todd/workspace/sid/src/resource/resource.c:434)
libsystemd.so.0!source_dispatch(sd_event_source * s) (/home/todd/workspace/systemd/src/libsystemd/sd-event/sd-event.c:3498)
libsystemd.so.0!sd_event_dispatch(sd_event * e) (/home/todd/workspace/systemd/src/libsystemd/sd-event/sd-event.c:3950)
libsystemd.so.0!sd_event_run(sd_event * e, uint64_t timeout) (/home/todd/workspace/systemd/src/libsystemd/sd-event/sd-event.c:4011)
libsystemd.so.0!sd_event_loop(sd_event * e) (/home/todd/workspace/systemd/src/libsystemd/sd-event/sd-event.c:4032)
libsidresource.so.0!sid_resource_run_event_loop(sid_resource_t * res) (/home/todd/workspace/sid/src/resource/resource.c:984)
main(int argc, char ** argv) (/home/todd/workspace/sid/src/daemon/sid.c:178)
```

We could raise(SIGTERM) in _on_worker_signal_event().  I was thinking it would be better to figure out what we are going to do long term with the worker rather than using raise()?

prajnoha:
> I think the child exiting is currently dependent on the parent reaching where the SIGTERM is sent in:
> 
> https://github.com/sid-project/sid/blob/93563cd39dfd686028da274e3f972f1660b5dba0/src/resource/worker-control.c#L319
> 
> We could raise(SIGTERM) in _on_worker_signal_event(). I was thinking it would be better to figure out what we are going to do long term with the worker rather than using raise()?

I see!

I think we can `raise(SIGTERM)`, such use in this case seems appropriate. But instead of `_on_worker_signal_event`, I would do that in `worker_control_worker_yield` where we can check if `worker->parent_exited == 1` and if yes, then the worker can terminate itself via `raise(SIGTERM)` (instead of calling `_chan_buf_send(chan, WORKER_CHANNEL_CMD_YIELD, NULL)` there).

I think this would be cleaner, because worker, in general, can still have some jobs to do in which case we shouldn't terminate it right after the "parent exited" signal. Of course, any communication with the parent would fail with that `-ENOLINK` which the worker can check when trying to use `worker_channel_send`, but at least there's still room for any cleanup needed or finishing up remaining jobs in the worker (which don't necessarily need to communicate with the parent).

(Obviously, the worker won't be reused anymore because there's no parent/manager that would sent it completely new jobs to process, so the only reason for delaying the `raise(SIGTERM)` would be that "finishing up any current jobs the worker has".)

trgill:
> > I think the child exiting is currently dependent on the parent reaching where the SIGTERM is sent in:
> > https://github.com/sid-project/sid/blob/93563cd39dfd686028da274e3f972f1660b5dba0/src/resource/worker-control.c#L319
> > 
> > We could raise(SIGTERM) in _on_worker_signal_event(). I was thinking it would be better to figure out what we are going to do long term with the worker rather than using raise()?
> 
> I see!
> 
> I think we can `raise(SIGTERM)`, such use in this case seems appropriate. But instead of `_on_worker_signal_event`, I would do that in `worker_control_worker_yield` where we can check if `worker->parent_exited == 1` and if yes, then the worker can terminate itself via `raise(SIGTERM)` (instead of calling `_chan_buf_send(chan, WORKER_CHANNEL_CMD_YIELD, NULL)` there).
> 
> I think this would be cleaner, because worker, in general, can still have some jobs to do in which case we shouldn't terminate it right after the "parent exited" signal. Of course, any communication with the parent would fail with that `-ENOLINK` which the worker can check when trying to use `worker_channel_send`, but at least there's still room for any cleanup needed or finishing up remaining jobs in the worker (which don't necessarily need to communicate with the parent).

Ok.  I implemented it as you described.  After testing - I don't see any problems.

What are you doing for formatting?  I see that .clang-format.  I typically use an IDE that does auto format on save - but that is formatting parts of the code that are not part of what I'm changing.  Should I open a issue to get all the source into the correct format?

> 
> (Obviously, the worker won't be reused anymore because there's no parent/manager that would sent it completely new jobs to process, so the only reason for delaying the `raise(SIGTERM)` would be that "finishing up any current jobs the worker has".)

I had been thinking that we should fork() a "worker manager" at startup.  But given that forking a process is so lightweight (COW etc), maybe se should just fork a process for each task?

prajnoha:
> What are you doing for formatting? I see that .clang-format. I typically use an IDE that does auto format on save - but that is formatting parts of the code that are not part of what I'm changing. Should I open a issue to get all the source into the correct format?

Hmm, which files are still showing as not formatted? I'm using a helper script `clang-format-all <top_level_directory>` (https://github.com/eklitzke/clang-format-all), but basically, it this is a `find` for all the `*.c` and `*.h` and then calling `clang-format` with the config on the files found.

> > (Obviously, the worker won't be reused anymore because there's no parent/manager that would sent it completely new jobs to process, so the only reason for delaying the `raise(SIGTERM)` would be that "finishing up any current jobs the worker has".)
> 
> I had been thinking that we should fork() a "worker manager" at startup. But given that forking a process is so lightweight (COW etc), maybe se should just fork a process for each task?

....I meant, the worker won't be reused because the parent ("worker manager") died for some reason.

As for forking a worker manager at startup to keep the pool of workers - right now, we're blocked by the fact that we're relying on that forking to do a db snapshot. But once we use a different way of db snapshotting scheme, then we can revisit this topic and weigh pros and cons and check how much loading time would pool of workers really save. If the initialization of the worker is kept lightweight, then we probably won't need that and we can just create a new process/worker for each incoming command.

prajnoha:
The patches look good, will merge them... just check whether the IDE consumes the clang-format, looks some bits are wrong, noticed the spaces and the "stars" before pointer variables are different. Thanks!

trgill:
> The patches look good, will merge them... just check whether the IDE consumes the clang-format, looks some bits are wrong, noticed the spaces and the "stars" before pointer variables are different. Thanks!

Ok.  Will do.

I'll update Jenkins to reject any PRs that don't conform to code standards while I'm at it.

trgill:
@prajnoha Jenkins is now updated to fail if code doesn't conform to clang-format.  Please ignore my comments about the IDE.  I had it configured wrong.

trgill:
@prajnoha I think I fixed the formatting problems and this is ready to merge.  Please let me know if you'd like changes.

prajnoha:
Looks good, will merge, thank you! (...I was away last week)

PR #101, opened 2021-04-27
kv export code merge

bmarzins:
This patchset makes _export_kv_store() use a memfd backed buffer, and uses the same code to build the kv dump for sidctl.

### DIFF
--- a/src/base/formatter.c
+++ b/src/base/formatter.c
@@ -30,10 +30,20 @@
 #define PRINT_JSON_END_ARRAY   "]"
 #define PRINT_JSON_INDENT      "    "
 
+static void _print_fmt(struct buffer *buf, const char *fmt, ...)
+{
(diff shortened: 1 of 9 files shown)


prajnoha:
All looking good, thanks!

As for records from UDEV namespace, maybe we could display the ones which are persistent+reserved so we can check easily which module reserves certain keys (like those various `ID_FS_*` for `blkid` and others). But that's just a detail, not that important, we can do that later as follow-up patch...

bmarzins:
> All looking good, thanks!
> 
> As for records from UDEV namespace, maybe we could display the ones which are persistent+reserved so we can check easily which module reserves certain keys (like those various `ID_FS_*` for `blkid` and others). But that's just a detail, not that important, we can do that later as follow-up patch...

Sure. I can change this.

PR #105, opened 2021-05-29
Iface changes

bmarzins:
These are changes to the interface between SID and its client programs, that are preliminary to making the client programs wrappers around a library interface. The biggest change is that the ouput formatting is now done by the SID daemon itself.  The client programs (sidctl, and usid) have been updated to work with the sid daemon doing the formatting, however no real work has been done to make them work more efficiently with these changes.  That will happen with the future changes adding a public library interface that they will use.


prajnoha:
Hey! Code looks good... As for the formatting itself, I've tried all the combinations of `sidctl <cmd> <--format>` and collected these:

  - `sidctl dump -f table` contains "siddb" as the very first line which is a bit superfluous I think for this format (if we kept it there, we could add a newline after it, if possible)

  - I'm thinking about `sidctl tree -f table` and `sidctl tree -f env` whether we can enahnce the output somehow - the tree contains these arrays like `children` and `event_sources` which makes it a bit harder to format that in table/env because we can't use usual "format blocks" like with JSON (with `{ }` and `[ ]`). For example:

```
❯ sidctl tree -f table
ID: worker/b319fab1-ac97-41e4-b2b0-37c9c4bf7599
type: worker
event-sources:
name: main

name: worker_signal_handler
...
```

So there's `event_sources` on one line and followed by two keys with the same name which makes it a bit confusing where one "block" (array) actually ends. I'm thinking whether it would be possible to add some delimiters there based on the nesting level or whether we're printing the array, don't know exactly what the best approach/representation would be here to make it a bit more readable/parsable... 

===

Also, the there's the extra copy in `sidctl` to output buffer - maybe we could just print the received data directly in those `_usid_cmd_<cmd_name>` functions instead of creating the `outbuf`, passing it to the `_usid_cmd_<cmd_name>`, copying to `outbuf` and printing the `outbuf` outside those functions.

prajnoha:
I've added some additional patches on top of yours that came to my mind when I saw the code... just a few cleanups that we can do now I think: https://github.com/prajnoha/sid/tree/ben-iface_changes. I've tried that only quickly, not checking it thoroughlly so glitches are possible :) (but output looked the same at first glance).

bmarzins:
> So there's `event_sources` on one line and followed by two keys with the same name which makes it a bit confusing where one "block" (array) actually ends. I'm thinking whether it would be possible to add some delimiters there based on the nesting level or whether we're printing the array, don't know exactly what the best approach/representation would be here to make it a bit more readable/parsable...

One thing I struggled with is that the ENV format is pretty much useless as ENV ouput for anything with arrays, since all the keys are the same for all the array elements. I'm not sure if we should disallow it, or what. Also, I wonder if different commands should have different default formats.  Or perhaps sidctl should just always use JSON by default, since that is always comprehensible.

I did tweak how spaces are used in the TABLE output to make it as readable as I could without making use of the indenting information in the print commands. But if we are willing to use the indenting information, we can make it more readable.

> 
> ===
> 
> Also, the there's the extra copy in `sidctl` to output buffer - maybe we could just print the received data directly in those `_usid_cmd_<cmd_name>` functions instead of creating the `outbuf`, passing it to the `_usid_cmd_<cmd_name>`, copying to `outbuf` and printing the `outbuf` outside those functions.

Yeah. Like I mentioned in my pull request message, I just got the client code functional.  There didn't seem much point in worrying about efficiency when I'm about to go rewrite that code to make it use a library.

bmarzins:
> I've added some additional patches on top of yours that came to my mind when I saw the code... just a few cleanups that we can do now I think: https://github.com/prajnoha/sid/tree/ben-iface_changes. I've tried that only quickly, not checking it thoroughly so glitches are possible :) (but output looked the same at first glance).

These look good. I have just one nitpick. Either exp_buf doesn't need to be part of the ucmd_ctx, or it doesn't need to be passed separately to _build_kv_buffer(), which is already has access to the ucmd_ctx.

prajnoha:
> One thing I struggled with is that the ENV format is pretty much useless as ENV ouput for anything with arrays, since all the keys are the same for all the array elements. I'm not sure if we should disallow it, or what. Also, I wonder if different commands should have different default formats. Or perhaps sidctl should just always use JSON by default, since that is always comprehensible.

The `-f table` looks good as default if we consider that there's human reading it and I'd probably keep that as default. The `-f env` vs. arrays - well, we'd need to do some kind of enumeration for the parent (array) key, like `event_source=source_name1 source_name2 source_name3 ...`. But that would require some further tweaking of the formatting code so I'd park that for now and we can revisit later. It's not that urgent.

> I did tweak how spaces are used in the TABLE output to make it as readable as I could without making use of the indenting information in the print commands. But if we are willing to use the indenting information, we can make it more readable.

Sure, I just noticed that extra newline in that `event_sources` array there in `sidctl tree -f table`. As for using indenting info - we can, but again, we can leave it for further fine tuning. I think for the API/easy parsing, the JSON is the important one for now.

> Yeah. Like I mentioned in my pull request message, I just got the client code functional. There didn't seem much point in worrying about efficiency when I'm about to go rewrite that code to make it use a library.

Yup...

> > https://github.com/prajnoha/sid/tree/ben-iface_changes
> These look good. I have just one nitpick. Either exp_buf doesn't need to be part of the ucmd_ctx, or it doesn't need to be passed separately to _build_kv_buffer(), which is already has access to the ucmd_ctx.

You're right! Looking at this part a bit more, I think we don't even need to have `res_buf` in the `ucmd_ctx` as it's not supposed to be used by module code directly (`ucmd_ctx` is supposed to be context for modules primarily). So I think we can clean this up a bit more and move the `res_buf` and `exp_buf` somewhere else (maybe passing that in the `exec_arg` would suffice). But will check and update...

prajnoha:
So I'll merge this and anything else we can do as follow-up patches, OK?

bmarzins:
Yep. If I'm really not ready for you to merge a pull request, I'll make it a draft pull request.

PR #106, opened 2021-06-05
Library interface

bmarzins:
This is beginnings of making the public library interface, starting from the existing libsidiface_usid library. There's clearly a lot of work left to be done. I just wanted to put this up early to make sure we all agree on the direction this is going.

My goal is to make it possible for callers to only need to link against one library, and for that library to only export symbols and define structures that are necessary for external programs to talk to the SID daemon. To this end, I pulled out all the code that makes use of buffers from the interface. However, internally the library still uses buffers and the comms_unix_* functions, and thus, users of the library still need to link against libsidbase.

To avoid this, I'd like to compile the comms_unix_* functions directly into the interface library. I don't have enough experience with autotools to know off the top of my head whether its easier to do this by making a static library  with the comms.c file for the inferface library to link against, or whether the inferface library should directly build with the comms.c file as a source.

Since the interface library doesn't need much of the buffers functionality, I was thinking of simply not using the buffer code. The library just needs to write the size of the data it will send, and then write that data, and then read the size of the data it will receive, allocate the necessary space, and then read the data. However we could also just compile the buffer code directly into the library, like I plan to do with the comms code.

I should point out that there currently is some memory inefficiency in this new interface, relative to the old one. The old usid_req() interface let callers pass in a function to populate the buffer that was being sent to the server  with extra data.  For ease of use, and to remove the need for callers to pull in the buffer code, this new interface just takes a pointer to some data to send. Currently this means that the user needs to allocate memory to store this extra data, and then the library will make a copy of that data to send the the SID daemon.  My current plan to avoid this is to have usid_req() build up the  extra data itself for commands (like scan and checkpoint) that have well defined data to send to SID. So for scan, the usid_req() callers wouldn't need to add any data, while for checkpoint, they would just need to add the name and keys to be sent. 

Thoughts?


prajnoha:
Hmm, honestly, I need to think this direction over a little bit...

The original thought here was to chunk the functionality into those smaller libraries so it's easy to share the code and at the same time only link against pieces that are needed - the `libsidbase` being really the basic low-level common code that actually all of the `sid` commands and other libs would use (unless it's really really simple piece of code not needing any of that functionality). So the direction with removing that link against `libsidbase` is a bit against that idea :)

Static code - might be - but we'd lose that "common shared code in memory" then and also updates can become less clear (part of the code will use the shared libs another part of the code would be linked statically against different version possibly). But, of course, we need to weigh the added complexity with linking against libs vs. straightfoward code.

Also for buffers, I was thinking about providing some more "runtime" insights in the future about their use by collecting some statistics about memory use (lowest, peak...). Also automatic deallocation, or possibly warnings if buffers are not deallocated properly when certain code/part is reached - so adding certain kind of safety (but it's true that this is probably more suitable for the daemon code rather than quick onetime commands).

But I'll think about your suggestion, because there area certainly things to consider...

(looking through the code right now)

prajnoha:
The interface itself looks good, though I'm still much more leaned to using the shared libs...

Thinking of the situation where someone wants to use the library - it should be sufficient to add the `libsidiface_usid` as direct dependency. Then, if the `libsidiface_usid` itself uses any other libraries, they should be loaded automatically (like the `libsidbase`).

Now, I've tried a simple program that uses `libsidiface_usid` and I've noticed that we were actually missing the internal dependency (missing `libsidbase`)!
```
❯ ldd /usr/local/lib/sid/libsidiface_usid.so
	linux-vdso.so.1 (0x00007ffc813d8000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f13521a4000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f1352385000)
```

This part was missing in the Makefile:

```
src/iface/Makefile.am | 2 ++                                                                                                                                  
 1 file changed, 2 insertions(+)
 
diff --git a/src/iface/Makefile.am b/src/iface/Makefile.am
index ec78b65..15f3b3c 100644
--- a/src/iface/Makefile.am
+++ b/src/iface/Makefile.am
@@ -36,6 +36,8 @@ libsidiface_servicelink_la_LIBADD = $(top_builddir)/src/base/libsidbase.la \
 
 libsidiface_usid_la_SOURCES = usid.c
 
+libsidiface_usid_la_LIBADD = $(top_builddir)/src/base/libsidbase.la
+
 iface_usiddir = ${ifacedir}
 
 iface_usid_HEADERS = $(top_builddir)/src/include/iface/usid.h
```

With this, it's just enough to call:

```
❯ gcc -I /usr/local/include/sid -L /usr/local/lib/sid -lsidiface_usid test.c
```

The `libsidbase` is added automatically through the internal dependency then:

```
❯ ldd /usr/local/lib/sid/libsidiface_usid.so
	linux-vdso.so.1 (0x00007ffc3b798000)
	libsidbase.so.0 => /usr/local/lib/sid/libsidbase.so.0 (0x00007fdfcaa64000)
	libuuid.so.1 => /lib64/libuuid.so.1 (0x00007fdfcaa50000)
	libc.so.6 => /lib64/libc.so.6 (0x00007fdfca881000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fdfcaa7a000)
```

```
❯ ldd a.out 
	linux-vdso.so.1 (0x00007ffdd45f7000)
	libsidiface_usid.so.0 => /usr/local/lib/sid/libsidiface_usid.so.0 (0x00007f6121c02000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f6121a28000)
	libsidbase.so.0 => /usr/local/lib/sid/libsidbase.so.0 (0x00007f6121a19000)
	libuuid.so.1 => /lib64/libuuid.so.1 (0x00007f6121a10000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f6121c09000)
```

prajnoha:
> I should point out that there currently is some memory inefficiency in this new interface, relative to the old one. The old usid_req() interface let callers pass in a function to populate the buffer that was being sent to the server with extra data. For ease of use, and to remove the need for callers to pull in the buffer code, this new interface just takes a pointer to some data to send. Currently this means that the user needs to allocate memory to store this extra data, and then the library will make a copy of that data to send the the SID daemon.

...for sending, if we used vector buffer, we could just glue the parts together composed of several sources and send that as request. But then the issue would be with reading the result because we're reusing the same buffer here we used for sending the request. And we don't support `buffer_read` for vector buffers (...that would allocate the memory needed to store the result). So we'd need to either use two buffers - one for sending request (vector) and one for reading result (linear) OR we'd need to add support for that `buffer_read` for vector buffers...

> My current plan to avoid this is to have usid_req() build up the extra data itself for commands (like scan and checkpoint) that have well defined data to send to SID. So for scan, the usid_req() callers wouldn't need to add any data, while for checkpoint, they would just need to add the name and keys to be sent.

Yeah, this probably sounds cleaner... But we need to count with the possibility that there might be more commands in the future with different extra data to add so the interface should probably just accept pointer to this extra data and then glue that to the request. Like the `data_fn` did in the original code, but without the need for API users to use the buffer code directly (if we don't find a better generic "extra argument/data accepting" interface).

bmarzins:
> Thinking of the situation where someone wants to use the library - it should be sufficient to add the `libsidiface_usid` as direct dependency. Then, if the `libsidiface_usid` itself uses any other libraries, they should be loaded automatically (like the `libsidbase`).
> 
> Now, I've tried a simple program that uses `libsidiface_usid` and I've noticed that we were actually missing the internal dependency (missing `libsidbase`)!

Here's my objection in a nutshell. The buffer functions are very generically named, so it seems totally possible that a user of this library will have functions with the same name.  If that happens, the user program will happily compile (well, right now usid.h includes buffer.h, so it might not compile if it sees both definitions of a buffer function at the same time. But there's no reason for buffer.h to be included in usid.h anymore, and I've removed it in my local repository). However, when the user program is run, libsidiface_usid will bind to the buffer functions that are locally defined by the user program, instead of the functions in libsidbase.  This will cause all sorts of problems that won't make any sense to the user. If you know some way around this, I'd have no objection to using the libsidbase. Otherwise, for safety's sake, it's better if the only functions that the public library exports are the ones that the user might want to use to talk to sid, which should all be in the .h file for the library.

bmarzins:
> > I should point out that there currently is some memory inefficiency in this new interface, relative to the old one. The old usid_req() interface let callers pass in a function to populate the buffer that was being sent to the server with extra data. For ease of use, and to remove the need for callers to pull in the buffer code, this new interface just takes a pointer to some data to send. Currently this means that the user needs to allocate memory to store this extra data, and then the library will make a copy of that data to send the the SID daemon.
> 
> ...for sending, if we used vector buffer, we could just glue the parts together composed of several sources and send that as request. But then the issue would be with reading the result because we're reusing the same buffer here we used for sending the request. And we don't support `buffer_read` for vector buffers (...that would allocate the memory needed to store the result). So we'd need to either use two buffers - one for sending request (vector) and one for reading result (linear) OR we'd need to add support for that `buffer_read` for vector buffers...

It seems pretty tricky to me to support `buffer_read` for vectors, without changing the interface to something more complicated, and adding a bunch of code just to support it.  The benefits don't seem to be worth it. One of the reasons I wanted to skip using the buffer interface at all in the library code was that it would allow us to just write to the `socket_fd` directly from the passed in memory, so there would be no need to make a copy.  But having `usid_req()` build up the extra data internally will fix the issue, while allowing it to continue to use the buffer interface internally.

> > My current plan to avoid this is to have usid_req() build up the extra data itself for commands (like scan and checkpoint) that have well defined data to send to SID. So for scan, the usid_req() callers wouldn't need to add any data, while for checkpoint, they would just need to add the name and keys to be sent.
> 
> Yeah, this probably sounds cleaner... But we need to count with the possibility that there might be more commands in the future with different extra data to add so the interface should probably just accept pointer to this extra data and then glue that to the request. Like the `data_fn` did in the original code, but without the need for API users to use the buffer code directly (if we don't find a better generic "extra argument/data accepting" interface).

I was going to leave the interface the same, so it would still include a `const void *data` and  `size_t data_len`. The `checkpoint` command currently needs them.  For some commands with more variable extra data requirements than `checkpoint` and `scan`, `usid_req()` would probably just directly append whatever data the user passed in. It's also possible that we might want to have a flag so that users can tell `usid_req()` that they don't want to use it's code to build the extra data internally, and it should simply append the passed in data, even for function where `usid_req()` would normally build up the extra data itself.

bmarzins:
Here's an example of the issue I was talking about with library users having a buffer_* function in their source, when libsidiface_usid links to libsidbase.
[https://gist.github.com/bmarzins/14ec891da987ea27f0593055d90e0f72](https://gist.github.com/bmarzins/14ec891da987ea27f0593055d90e0f72)

prajnoha:
> Here's my objection in a nutshell. The buffer functions are very generically named, so it seems totally possible that a user of this library will have functions with the same name.

> However, when the user program is run, libsidiface_usid will bind to the buffer functions that are locally defined by the user program, instead of the functions in libsidbase.

Yes, that a pain point, I understand. The names are not ideal for a library. I wanted to rename all that is in those library interfaces some time ago, adding proper prefixes, but didn't get to that in the end, it flew out of my head completely :( But I think we can still do that without any harm as the libs are not practically used anywhere yet but our own code. Namespaces would be better, but we have C here...

I wonder if adding explicit exported symbols file (the linker's `--version-script` option) would help. libdm uses that in a way for symbol versioning and I've seen that systemd lib has one too. If we took the `libsidiface_usid` for example, that exports file would list only the items directly related to that library and any internally used library symbols should be hidden then.

But then I'm not 100% sure how it would work if, for example, some code needed both the `libsidiface_usid` as well as `libsidbase` (...I mean `libsidbase` not being only internal dependency for `libsidiface_usid`, but both libraries directly used in a code and similar situations).

prajnoha:
> It seems pretty tricky to me to support `buffer_read` for vectors, without changing the interface to something more complicated, and adding a bunch of code just to support it. The benefits don't seem to be worth it.

Sure, I agree...

> One of the reasons I wanted to skip using the buffer interface at all in the library code was that it would allow us to just write to the `socket_fd` directly from the passed in memory, so there would be no need to make a copy. But having `usid_req()` build up the extra data internally will fix the issue, while allowing it to continue to use the buffer interface internally.
> 

As you said, right now, it's only that `scan` and `checkpoint` that needs to append some data to the buffer. So we can probably leave the buffer in place and if there's a need in the future to add more custom data, we can think of alternatives - if the buffer stays hidden behind the interface, we can replace that later either by better/enhanced buffer implementation or simply removing the buffer underneath and using direct approach instead. But such change would stay transparent for library users.

> > > My current plan to avoid this is to have usid_req() build up the extra data itself for commands (like scan and checkpoint) that have well defined data to send to SID. So for scan, the usid_req() callers wouldn't need to add any data, while for checkpoint, they would just need to add the name and keys to be sent.

> I was going to leave the interface the same, so it would still include a `const void *data` and `size_t data_len`. The `checkpoint` command currently needs them. For some commands with more variable extra data requirements than `checkpoint` and `scan`, `usid_req()` would probably just directly append whatever data the user passed in. It's also possible that we might want to have a flag so that users can tell `usid_req()` that they don't want to use it's code to build the extra data internally, and it should simply append the passed in data, even for function where `usid_req()` would normally build up the extra data itself.

Yup, we can do that. Maybe we could also make `usid_req` to accept an argument that is a structure/union and then based on the type, it could select the appropriate field, for example:

```
struct usid_request {
  usid_cmd_t cmd;
  uint16_t flags;
  uint64_t status;
  size_t extra_data_len;
  union extra_data {
     data_for_cmdA;
     data_for_cmdB;
     ...
  };
};
```

or if we don't expect that every `extra_data` is of variable size, we can just specify the size where needed:

```
/* data for cmdA that has all static sized elements */
struct data_for_cmdA {
  field A;
  field B;
  ...
};

/* data for cmdB that has one variable-sized element */
struct data_for_cmdB {
  size_t fieldA_size;
  field *A;
  field B;
  ...
};

struct usid_request {
  usid_cmd_t cmd;
  uint16_t flags;
  uint64_t status;
  union extra_data {
     struct data_for_cmdA cmdA;
     struct data_for_cmdB cmdB;
     ...
  };
};
```

Then the `usid_req` would look like:

```
int usid_req(struct usid_request *req, struct usid_result **res_p);
```

That way, it would enforce library users  to think and always realize a bit what they're actually passing in, so we don't have the generic `void *data` element there that could be just anything and it's potential source of problems associated with passing improper type of data. What do you think?

prajnoha:
> That way, it would enforce library users to think and always realize a bit what they're actually passing in, so we don't have the generic `void *data` element there that could be just anything and it's potential source of problems associated with passing improper type of data. Of course, it's a bit more laborious to construct the structure compared to passing the `void *`, but that's the point - to cause users stop here for a while and realize what is passed in. What do you think?

...actually, I'm using similar approach here already, which I also found that it makes debugging more clearer (seeing directly the structure instead of generic `void *` pointers):

https://github.com/sid-project/sid/blob/4c46c37a301872a28ac19e905f260b8116ce66de/src/include/resource/worker-control.h#L64-L88

bmarzins:
So, I tried playing around with using a version script, and while it's easy to use it to limit exporting symbols, AFAICT it doesn't do anything to force the library to bind its undefined symbols to ones in its dependent libraries, instead of to ones in a program that depends on it.

bmarzins:
> Well, yes, for this particular issue we're left with proper naming only (adding prefixes) to avoid the collisions as much as possible. But this is an issue that we can induce anywhere - libraries do use other libraries (e.g. I can make libdevmapper to call my own `udev_new` or systemd to call whatever it calls from other libs...).

The difference is that libudev is itself a public library with a documented, stable API, while libsidbase isn't really.
 
> I would like to avoid "special cases" where in one case we'd be using a shared lib and in the other we'd using statically attached code - it might get confusing later and can cause maintenance pain. The more we can reuse the same code (in same manner), the better, I think... Even though it has its pitfalls like you mention - not sure if there's any trick we could use to avoid this problem, but if there was one, we could consider it (...but I'm afraid we're left with that proper naming).

I suppose we could start treating libsidbase as something closer to a public library.  Certainly, adding prefixes would go a long way. I'm not sure off the top of my head if there are some exported symbols in libsidbase that are only used in files that are also in libsidbase.  If so, then a version script could help to make sure that we are only exporting symbols that are needed outside the library.

bmarzins:
Speaking of changing names, do you think we should keep the library name as libsidiface_usid, and the prefix as usid? It will obviously be used by usid, but I suspect that this will be the only official library interface to sid, so perhaps libsidiface or just libsid makes more sense.

prajnoha:
> The difference is that libudev is itself a public library with a documented, stable API, while libsidbase isn't really.

True.

> I suppose we could start treating libsidbase as something closer to a public library. Certainly, adding prefixes would go a long way.

Yes, making that public was my intention (though later in the future). OK, so let's rename then. The libs are still internal, so we're not messing anything external yet.

Right now, we have these header files:

```
./base:
base64.h  bitmap.h  buffer-common.h  buffer.h  buffer-type.h  common.h  comms.h  formatter.h  hash.h  list.h  mem.h  util.h

./iface:
service-link.h  usid.h

./log:
log.h

./resource:
kv-store.h  module.h  module-registry.h  resource.h  resource-type-regs.h  ucmd-module.h  worker-control.h
```

The `resource` part already has the `sid_` prefix, so "only" the remaining bits need renaming. Will be a huge change because mainly the base lib is used all over the code, the code formatting will also produce changes, but that's inevitable...

The `resource` is supposed to be the library providing higher-level code that reuses other parts and creates reusable units.

I've checked the `log` lib and noticed there was superfluous dependency on `base` lib - I've removed that in 93563cd39dfd686028da274e3f972f1660b5dba0. The point of log lib being separate is to be able to use pure logging facility without bringing in the base lib (or others) if not necessary.

The `iface` has two interfaces right now - the `service-link.h` and `usid.h` and separated into two distinct libs (`libsidiface_usid` and `libsidiface_servicelink`). The original point here was that SID could have different interfaces to use, either interfacing the SID itself from outside (like the `usid`) or interfacing from inside the SID to external libs/tools (like the `servicelink`). And that SID will contain small helper tools to access various parts of the SID functionality.

The `libsidiface_usid` was supposed to handle the sid <--> udev communication currently handled by our own `usid` command, but possibly reused for `usid` udev builtin command in the future.

The `libsidiface_servicelink` is certainly something that can be reused in those other possible helper tools, if they ever need to be executed as a service.

Then my plan was to have something like `libsidiface_db` to access SID's database with another helper tool like `siddb`. But still making it possible to access the db directly through a library that other tools can make use of and possibly add better presentation for the records, GUIs, whatever... (so `siddb` would be just the simple one distributed by us and letting others to add their own, more advanced db presentation).

So that is the reason behind that code separation in libs - mainly the code reuse either by our own small helper tools or external libs/tools for futher enhancing/reuse.

But it's true that we have the `sidctl` now which also has the originally meant `siddb` functionality too in a way (though it's a full dump feature only). So maybe more advanced features like record selection/filtering/streaming would be a part of that more focused `siddb` tool (reusing the originally considered `libsidiface_db`).

This is how I meant it to be originally - small reusable parts by both our own code and external code and to make it easy to enhance. So that's why we have the `libsidiface_usid` - because it was meant to be used only for our own `usid` cmd or `usid` udev builtin cmd (and has to be minimal lib so that external tools/libs have less reasons to fuss about various the lib dependencies and/or lib size when proposing to link against such library, making it easier for them to accept).

bmarzins:
> This is how I meant it to be originally - small reusable parts by both our own code and external code and to make it easy to enhance. So that's why we have the `libsidiface_usid` - because it was meant to be used only for our own `usid` cmd or `usid` udev builtin cmd (and has to be minimal lib so that external tools/libs have less reasons to fuss about various lib dependencies and/or lib size when we try to propose to link against such library, making it easier for them to accept).

That all makes sense, but at least right now usid_req() is the interface for sending any request to the SID daemon, and getting back the result.  It's possible that convenience functions will be layered on top for some future commands, potentially in a different library, but I'm pretty sure that everything will go through usid_req() to do the actual interfacing with the daemon.  If you're fine with the current library name, we can leave it as is, but it seems like a the current name is both long and potentially confusing, and something like libsid, or libsid_iface (with the functions all being prefixed with sid_, instead of usid_) would make more sense.

prajnoha:
> but I'm pretty sure that everything will go through usid_req() to do the actual interfacing with the daemon. If you're fine with the current library name, we can leave it as is, but it seems like a the current name is both long and potentially confusing, and something like libsid, or libsid_iface (with the functions all being prefixed with sid_, instead of usid_) would make more sense.

Well, OK, if it's really generic functionality like `usid_req`, then yes (...and yes, in that case, it should be just `sid_` prefix, not `usid_`).

But with the new changes, we're actually hardwiring the `USID_CMD_SCAN` and `USID_CMD_CHECKPOINT` extra data handling (which is specific to `usid` command) directly into this `usid_req` function, which kind of contradicts the intention to make it generic... If we keep going with this, it means that after adding any new command, we'd need to add that command number/identifier to the `usid_cmd_t` enum (or `sid_cmd_t` enum if we're going to leave the "u" in the prefix) all the time. Then we'd probably need some way of checking whether the caller is authorized to call such command - e.g. if it's a library to access only the db, we'd allow only db-specific commands/functionality, but we'd not allow commands like `scan`, `checkpoint` etc...

prajnoha:
But I think such "global" list of commands is OK - the main "sid" library would contain all of them, but then the higher level libraries that handle specific functionality (like db, sidctl, usid...) would only use a subset of that full command set and they would wrap the low-level `sid_req` call. So yes, I think that could work...

bmarzins:
I think it's probably worth reviewing and pulling these changes now. There is still some work that needs to be done, such as adding comments to iface.h and possibly adding a man page. There's also the grand prefixing that needs to happen for libraries other than libsidiface. But those changes can be in a future pull request.

bmarzins:
For the next step, I've been thinking that, instead of prefixing all the functions in libsidbase, it might be better to split that library into two, say libsidbase, and libsidinternal, where libsidbase has the things that the documented library interfaces need, like the buffer and comms code, and the libsidinternal has things that people aren't going to expect to get from a public sid library, like the bitmap, list, and hash code. Programs like sidctl and usid could possibly use these, but I don't think that many outside users would link to  libsidbase to get this non-sid related code. Since the libsidiface code doesn't need these, they could be kept for use in the SID package, and we don't need to worry about having to sid_ prefix all those functions to avoid clashing with hash, list, or bitmap functions that users of libsidbase have.  If we find that we need to move code from libsidinternal to libsidbase latter, that should be fine.

We should probably also make the base64 code part of a documented public interface, so that users can decode
the encoded binary output the SID generates.

prajnoha:
> I think it's probably worth reviewing and pulling these changes now. There is still some work that needs to be done, such as adding comments to iface.h and possibly adding a man page. There's also the grand prefixing that needs to happen for libraries other than libsidiface. But those changes can be in a future pull request.

Yup, the patches look good, will merge... Thanks!

> For the next step, I've been thinking that, instead of prefixing all the functions in libsidbase, it might be better to split that library into two, say libsidbase, and libsidinternal, where libsidbase has the things that the documented library interfaces need, like the buffer and comms code, and the libsidinternal has things that people aren't going to expect to get from a public sid library, like the bitmap, list, and hash code. Programs like sidctl and usid could possibly use these, but I don't think that many outside users would link to libsidbase to get this non-sid related code. Since the libsidiface code doesn't need these, they could be kept for use in the SID package, and we don't need to worry about having to sid_ prefix all those functions to avoid clashing with hash, list, or bitmap functions that users of libsidbase have. If we find that we need to move code from libsidinternal to libsidbase latter, that should be fine.

OK, that sounds reasonable... If we ever needed some of the other functions "outside", we could just move from internal to public lib (and renaming then). It's probably better to keep the public part minimal so we can still afford changes in the internal lib if needed without any further complications...

> We should probably also make the base64 code part of a documented public interface, so that users can decode
> the encoded binary output the SID generates.

Yup.

PR #113, opened 2021-07-02
Library cleanup

bmarzins:
This is the follow-up library work from my last pull request. The main changes are splitting out libsidinternal from libsidbase and prefixing the functions in libsidbase, so that there's less chance of a symbol conflicts.


blivetghoul:
Can one of the admins verify this patch?

trgill:
ok to test

trgill:
The Jenkins admin list somehow was cleared.  I fixed it.  Our PRs should be tested automatically now.

prajnoha (changes requested):
Looks fine! I'm just thinking whether we should be also renaming the structures, constants and typedefs when we're at this, just to minimize possible conflicts (...like the `struct buffer`, `buffer_pos_t` etc.).

I've noticed one of the tests fail, Jenkins with it then. It's just missing one dep on the new internal lib:

```
 src/iface/Makefile.am | 3 ++-
 1 file changed, 2 insertions(+), 1 deletion(-)

diff --git a/src/iface/Makefile.am b/src/iface/Makefile.am
index b16b0e8..8dc0db6 100644
--- a/src/iface/Makefile.am
+++ b/src/iface/Makefile.am
@@ -31,7 +31,8 @@ libsidiface_servicelink_la_CFLAGS = $(SYSTEMD_CFLAGS)
 
 libsidiface_servicelink_la_LDFLAGS = -version-info 0:0:0
 
-libsidiface_servicelink_la_LIBADD = $(top_builddir)/src/base/libsidbase.la \
+libsidiface_servicelink_la_LIBADD = $(top_builddir)/src/internal/libsidinternal.la \
+				    $(top_builddir)/src/base/libsidbase.la \
 				    $(SYSTEMD_LIBS)
 
 libsidiface_la_SOURCES = iface.c
```

Otherwise, looks good, thanks!

bmarzins:
Oops! Yeah. Forgot to push that.

bmarzins:
I'm not sure about prefixing the other parts of the header files. If libsidiface is the only real public interface that we want to support, then the only thing we need to worry about are the exported symbols for libsidbase. But like I mentioned before, we need to at least support the sid_binary_* functions for decoding the encoded binary. However, if we want to have all the libsidbase includes completely prefixed, I'm not sure what to do about config.h. Perhaps we should remove the dependencies on it from the rest of libsidbase.

prajnoha:
I think other parts besides `sid_binary_*` may come in handy too, like buffer which is quite versatile, also I can imagine the comms code to be reused, various utility/helper functions for parts which we find commonly used (e.g. logging, executing external commands, parsing the output, parsing key-value pairs and importing them to SID db, whatever...).

We need to take into account that we have modules for both the scanning as well as trigger/action part. Right now, these modules are in SID repo, but I expect they'll become a part of those other subsystem's repositories instead and just linking against SID's libs. So we won't be taking care of the exact module code, logic and all its internal intricacies, just providing the infrastructure to run them AND providing a set of helper functions for common tasks/code.

So from this perspective, I though that generic names like `struct buffer` are not quite appropriate, I have to admit, because:

  - it's not clear right from the name that such structure comes from a SID helper lib
 
  - and I can imagine someone simply trying to use the same name (of the structure, constant...) and then getting into the conflict unintentionally (of course, the compiler would fuss about two/more definitions of the same structure so it's easily fixable by lib user - just renaming his structure, but it's annoyance still which we can probably avoid)

prajnoha:
Hmmm, looking at the public interface for modules (the `src/include/ucmd-module.h`), I see it includes `internal/formatter.h` - we can remove that formatter include there, right?

prajnoha:
A nice pile of renaming, but we had to do that once :) Thanks a lot for cleaning it up! Just a few things I've noticed related to clang-format:

  - we have removed `base/common.h`, so we can remove that from `.clang-format` too
  - if we're including `config.h` in the source, I'd probably give it highest priority in `.clang-format`'s `IncludeCategories` setting so it's always at the very top, because it is the main source config and it can possibly change behavior of other includes/defines/whatever. So I'd probably go with this:
    -  prio0 --> `"config.h"`
    - prio1 --> `"internal/common.h"`
    - prio2 --> all local includes (the ones with `" "`)
    - prio3 --> the rest of includes (system ones with `< >`) except `<cmocka.h>`
    - prio4 --> `<cmocka.h>` (cmocka needs all system includes to be included first)
  - then please rerun the clang-format (when I try to run even without changing the IncludeCategories, it shows me some differences - I'd rerun that, but probably better if you do that per patch so it's not messy in git history then)

Thanks.

prajnoha (approved):

bmarzins:
It appears that if I make internal/common.h have priority 1, then the .h file with the same name as the .c file gets moved in front of it. So, I just made both config.h and internal/common.h have a priority of 0. Since internal/common.h includes config.h, there's no point to having both in the same .c file, and even if they both are, the ordering doesn't matter.

prajnoha:
> Since internal/common.h includes config.h, there's no point to having both in the same .c file, and even if they both are, the ordering doesn't matter.

...yup, sure, that's OK.

PR #116, opened 2021-08-16
add tests for kv-store

trgill:


src/resource/kv-store.c
```diff
@@ -363,7 +363,7 @@ static int _hash_update_fn(const char *               key,
 void *kv_store_set_value(sid_resource_t *          kv_store_res,
                          const char *              key,
                          void *                    value,
-                         size_t                    value_size,
+                         size_t                    value_count,
```
trgill, line 366: @prajnoha maybe  there a context where this is a size and not a count?

tests/test_kv_store.c
```diff
@@ -230,6 +231,48 @@ static void test_kvstore_iterate(void **state)
...
+	                                        KV_STORE_VALUE_OP_MERGE,
+	                                        _kv_write_always,
+	                                        &update_arg),
+	                     NULL);
+	assert_int_equal(kv_store_num_etries(kv_store_res), 1);
+	kv_return_value = kv_store_get_value(kv_store_res, MERGE_KEY, &kv_data_size, NULL);
+
+	assert_int_equal(test_data_size, kv_data_size);
+	assert_int_equal(kv_return_value->size, kv_data_size);
+	assert_int_equal(memcmp(kv_return_value->data, test_iov, kv_data_size), 0);
```
trgill, line 298: @prajnoha the test is failing on the memcmp().  when I step through the hash lookup all looks fine.  I think it is related to not using a KV_STORE_VALUE_REF | KV_STORE_VALUE_VECTOR parameter.  Is it obvious to you what I'm doing wrong?

tests/test_kv_store.c
```diff
@@ -232,6 +233,66 @@ static void test_kvstore_iterate(void **state)
...
+	test_data_size = add_test_data(test_iov);
+
+	assert_ptr_not_equal(kv_store_set_value(kv_store_res,
+	                                        MERGE_KEY,
+	                                        test_iov,
+	                                        MAX_TEST_ENTRIES,
+	                                        KV_STORE_VALUE_VECTOR,
+	                                        KV_STORE_VALUE_OP_MERGE,
+	                                        NULL,
+	                                        NULL),
```
trgill, line 277: @prajnoha if I called the function supplying the last two parameters:

```
	assert_ptr_not_equal(kv_store_set_value(kv_store_res,
	                                        TEST_KEY,
	                                        test_iov,
	                                        MAX_TEST_ENTRIES,
	                                        KV_STORE_VALUE_VECTOR,
	                                        KV_STORE_VALUE_NO_OP,
	                                        test_kv_write_always,
	                                        &update_arg),
```
The data validation failed.  It only happens with OP_MERGE.

I'll look at why it fails.

src/resource/kv-store.c
prajnoha: This is meant to be "vector item count" in case we have a vector (value flagged as `KV_STORE_VALUE_VECTOR`) - then there's size inside the `struct iovec` and its`iov_len` field for each vector item.

It's "real size" in case it's the usual scalar value (so not flagged as `KV_STORE_VALUE_VECTOR`). I used "size" for the name of the parameter which seemed to be more generic and could incorporate both vector and non-vector case...

trgill: I see.  I'll drop the commit.

tests/test_kv_store.c
trgill: I was mistaken about how KV_STORE_VALUE_OP_MERGE works.  I see how it works now.

trgill:
@prajnoha I think this is ready for review.

src/resource/kv-store.c
```diff
@@ -546,6 +546,12 @@ void kv_store_iter_destroy(kv_store_iter_t *iter)
 	free(iter);
 }
 
+size_t kv_store_num_etries(sid_resource_t *kv_store_res)
```
prajnoha, line 549: little typo: `etries` --> `entries`

tests/test_kv_store.c
```diff
@@ -60,11 +102,77 @@ static void test_type_H(void **state)
...
+
+	kv_store_res = sid_resource_create(SID_RESOURCE_NO_PARENT,
+	                                   &sid_resource_type_kv_store,
+	                                   SID_RESOURCE_RESTRICT_WALK_UP,
+	                                   "testkvstore",
+	                                   &main_kv_store_res_params,
+	                                   SID_RESOURCE_PRIO_NORMAL,
+	                                   SID_RESOURCE_NO_SERVICE_LINKS);
+
+	KV_VALUE_PREPARE_HEADER(test_iov, seqnum, flags, (char *) TEST_OWNER);
```
prajnoha, line 125: Here, for the `KV_VALUE_PREPARE_HEADER`, the flags passed are supposed to be of `sid_ucmd_kv_flags_t` type - these are the flags that define permissions and persistency (protected/private/reserved...) - this is the extra part that ubridge adds on top of the pure `kv_store_value`.

So it's the `struct kv_store_value` that `kv_store.c` defines:

  [size] [int_flags] [ext_flags] [kv_store_data]

Then the `data` from the `kv_store_value` above is further enriched in `ubridge.c` as `struct kv_value`:

  [seqnum] [flags] [data]

So the overall record is then actually:
  

  [size] [int_flags] [ext_flags] [kv_store_data == [seqnum] [flags] [data]]

tests/test_kv_store.c
```diff
@@ -60,11 +102,77 @@ static void test_type_H(void **state)
...
+	                                        KV_STORE_VALUE_VECTOR,
+	                                        KV_STORE_VALUE_NO_OP,
+	                                        _kv_overwrite,
+	                                        &update_arg),
+	                     NULL);
+
+	assert_ptr_not_equal(iter = kv_store_iter_create(kv_store_res), NULL);
+	/* validate the contents of the kv store */
+	while ((return_iov = kv_store_iter_next(iter, &data_size, &flags))) {
+		assert_true(flags & (KV_STORE_VALUE_REF | KV_STORE_VALUE_VECTOR));
```
prajnoha, line 149: ...so, if you called `kv_store_set_value(kv_store_res, ..., KV_STORE_VALUE_VECTOR, ...)`, the flags tested should be only `KV_STORE_VALUE_VECTOR` (without `KV_STORE_VALUE_REF`).

This test line passes now, because `flags & (KV_STORE_VALUE_REF | KV_STORE_VALUE_VECTOR` is true if either `KV_STORE_VALUE_REF` is set or `KV_STORE_VALUE_VECTOR` is set or both. So we need to check the exact flags here with `==` instead of `&` I think (and without `KV_STORE_VALUE_REF` if we're not using it for `kv_store_set_value`)

prajnoha (changes requested):
Hey! Thanks! Just a few minor things to fix, but otherwise looks good...

src/resource/kv-store.c
trgill: thnaks

trgill:
@prajnoha I think I've addressed your comments.  Please let me know if I missed something.

prajnoha:
Merged, thanks!

PR #120, opened 2022-02-21
hash replacement tkrzw

trgill:
prototype to replace the hash table with tkrzw


prajnoha:
Hi! I've been also looking at and trying `tkrzw` myself and I have collected a few notes so far when looking more into this, looking at details and even directly at the `tkrzw` code if I didn't find an answer directly in the documentation.  They are things we should consider, be aware of and probably discuss a bit more before really delving into `tkrzw`:

---
1)
When we're getting a value from db (`tkrzw_dbm_get`, `tkrzw_dbm_get_multi`, `tkrzw_dbm_iter_get_value` ...), the **returned value is always dynamically allocated** (`malloc`). We never get direct value from the db. So there's little overhead here, e.g. if I just want to read the value and do a quick decision based on it and then not needing it anymore within current processing.

If we wanted  to avoid this to certain degree, we could store just pointers to the values in the database - then only the memory to keep the pointer would be allocated. BUT that'll mean the 'dump/sync to a file' feature provided by the db won't work correctly because, well, the values would be just pointers and we'd need to dereference them first somehow. In summary - either we'll live with the extra `malloc` each time we get a value or we'll never use the db's sync to file feature.

Though, the reason why the db does the `malloc` for each returned value is understandable. It's in-memory database and as such, it needs to count with the situation that there might be `tkrzw_dbm_set` (or any other "set" which changes the record) right after we get the value and so we need to work with the copy, not the direct value from db. BUT in our case, we don't have any parallel access - we're working with the whole db snapshot in the worker process and there the access to db is always sequential. The same applies when we're syncing db changes back from worker process to main process.

---
2)
`tkrzw` claims it supports atomicity. Well, yes, it does take the proper locks with `tkrzw_dbm_get` and `tkrzw_dbm_get_multi` call so the transaction is applied as a whole... until there's no error hit. **If there's an error hit in the middle of the transaction, like `malloc` failure inside, there's no automatic db rollback to the point right before the transaction started**. This is something I'm a bit disappointed about. I'd expect that when atomicity is claimed to be supported, then either the transaction is applied as a whole or not, meaning, in any error case, the automatic rollback would happen. This is something I'm eagerly looking for.

Without the automatic rollback support, we have two options when dealing with transaction errors:

  - Either we completely throw whole database away and we recreate it (this is not stopper issue for SID through, it's just we'd need to reevaluate everything - that practically means running `udevadm trigger` - it's just clumsy).

  - Or we add a db log ourselves so we could do the rollback ourselves on top of the db.

---
3)
`tkrzw` has support for custom callbacks before the value is actually stored - the `tkrzw_dbm_process` and `tkrzw_dbm_process_multi` calls support this. That is very nice because this allows us to map our `kv_store_update_fn_t` type of callback we support in our `kv-store` wrapper.

Why we'd like to map directly `kv_store_update_fn_t` even in the backend db behind the wrapper? Because if we don't, we'll need to implement our `kv_store_set_value` like this:

  - get value from backend
  - call the `kv_store_update_fn_t` callback 
  - set value in backend

So we have `get` and `set` for the backend each time we call `kv_store_update_fn_t`. So first time, the backend looks up value for a key to get a value, the second time the backend looks up the key again. That means, always having two completely identical searches through the whole structure per one "set" operation, the only difference being if there was no callback defined for `kv_store_set_value`.

Why do we need the `kv_store_update_fn_t` callback at all? Because we need to see the previous value and act based on that, like adding/removing items to/from a vector value, checking permissions or key reservations before writing the value etc. This is the core functionality inside SID it's built on. We depend on this. The `kv_store_update_fn_t` is direct optimization to avoid the double lookup in the db structure.

Said that and that the `tkrzw` supports the callbacks, it's limited in one detail - we can't do any further db access inside the callback, including `tkrzw_dbm_get` and `tkrzw_dbm_set` call. This is understandable for the `set` where  we don't want to recursively change records. But we can't call `get` either. I've tried this, calling `tkrzw_dbm_get` inside the callback and it blocked.

Unfortunately, we do have situations where we'd like to collect/get more records while processing another record and we don't know the keys in advance. For example, we're trying to modify a value for key A, a callback is called which in turn gets the values for related keys B, C, ... (the set of these keys depends on the value of A actually), calculates the new value of A based on that and returns that. In these cases, we can't use the callback support inside the db backend and we're back to this sequence:

  - get value for key A
  - get list of related keys from the value of A
  - get all the value of related keys B, C ...
  - calculate the new value for key A
  - set the new value for key A

With tkrzw, we can't put this inside a transaction. Because for a transaction, we need to know the keys in advance.

4)
---
`tkrzw` is quite young project (starting in 2020), even though it's a successor to well-known Tokyo and Kyoto cabinet. Practically, this means the adoption in various distros is not yet complete. As for Fedora/RHEL, the package is in Fedora, but not in RHEL yet. In Debian, this is still in "unstable" branch. I haven't checked others. Of course, this is solvable, but it's something we need to keep in mind too.

---
5)
Considering the notes above and mapping to our needs:

A) *No need for locking actually to avoid mutual access - we always do a single operation at a time.*
It looks that `tkrzw` always uses locking, can't be disabled. Also, the returned values are always freshly allocated/copied, we never get direct values as stored in the db. When doing a single operation at a time, this is useless extra job.

B) *Need for atomic update of a transaction with automatic rollback support (so the transaction applied all or return to the original state).*
Only partially - `tkrzw` supports transactions with `multi` calls, but not automatic rollback support. There's a mention in the documentation that if using a file, there's some support for a rollback, but that is with "append" mode where the sice can go wild... I just need "append" mode for current transaction, not always appending.

C) *Need for callback support on 'set' and it would be great if 'get' could be called inside the callback (...but this is related to locking, if locking is not there, there's nothing to stop us from doing that).*
Only partially - `tkrzw` supports callbacks on "set", but we can't do any other "get" inside the callback... Again, we don't need to consider parallel access at the same time so this restriction comes from the fact the db tries to handle parallel access with locking.

D) *Need for ordered structure so we can retrieve records for key ranges and key prefixes.*
The b+ tree structure that `tkrzw` supports is perfect match for this.

---

In summary, only D) is fully supported and the way I envision. It all boils down to a need for a key-value storage structure supporting ordered keys for the range/prefix lookups - like a B+ tree. So we really need to consider this and the notes above. I even tried to find a library we could simply reuse for the B+ tree, but with no success yet. We might as well end up writing our own implementation - it's not something that would be awfully hard to do, it's just about having a correct and completely debugged version.

I'm probably giving up looking for more databases out there because all have some nice bits and then fall in other areas. So either we try to use the `tkrzw`, but probably more for its ordered structure for key prefix/range lookups than for its real database features. It's hard to find a good and complete match.

What do you think?

prajnoha:
We are not adding the `tkrzw` KV store backend right now - we switched to btree instead. Archiving this PR, we can revisit later in the future still...

PR #122, opened 2022-03-10
Miscellaneous fixups

### LINKED ISSUE #121: problems with how util_str_iterate_tokens()  deals with quotes
I can see two problems with how util_str_iterate_tokens() deals with quotes, using UTIL_STR_DEFAULT_QUOTES and UTIL_STR_DEFAULT_DELIMS.

1. Quotes aren't delimiters, so if the function was processing `can't or won't`, it would find 3 tokens: `can't`, `t or won`, and `t`.  Since the quote isn't a delimiter, the first token will keep reading right past it. For reference, if `can't or won't` was part of a bash command line, it would become the single argument `cant or wont`.

2. It checks for escape characters, but doesn't remove the escapes. So, there is no way to get a token like `"quoted string"`. You could either get `quoted string` or `\"quoted string\"`. You also can't get the token `can't or won't`. 

The way I would suggest to solve this would be to change how quotes are dealt with, in three ways.
1. There is no way to escape a quote character, except by placing it in a quoted token, delimited by a different quote character.
2. If there is no delimiter before the start of a quoted token, it is appended on to the previous token.
3. If there is no delimiter after a quoted token, the next token is appended on to it.

With these changes, you could get the token `"quoted string"` by passing in `'"quoted string"'` and `can't or won't` by passing in `can"'"t or won"'"t`. Also, processing `can't or won't` would get you one token, `cant or wont`, just like bash would.

I'm pretty sure this change makes sense, but since this would only get called with the default delims and quotes for an external worker, and there's currently no way to execute an external worker, I'm not exactly sure how the command arguments and environment variables will get passed. So, perhaps I'm missing something.

bmarzins:
These are some issues I noticed while coming back up to speed with the code.

Resolves #121

### DIFF
--- a/src/base/comms.c
+++ b/src/base/comms.c
@@ -29,7 +29,7 @@ static void _comms_unix_addr_init(const char *path, size_t path_len, struct sock
 {
 	memset(addr, 0, sizeof(*addr));
 	addr->sun_family = AF_UNIX;
-	memccpy(addr->sun_path, path, path_len, sizeof(addr->sun_path));
+	memcpy(addr->sun_path, path, path_len);
 	*addr_len = offsetof(struct sockaddr_un, sun_path) + path_len;
 }
 
@@ -40,7 +40,7 @@ int sid_comms_unix_create(const char *path, size_t path_len, int type)
 	socklen_t          addr_len;
 	int                r;
 
-	if (!path_len) {
+	if (!path_len || path_len >= sizeof(addr.sun_path)) {
 		r = -EINVAL;
 		goto fail;
 	}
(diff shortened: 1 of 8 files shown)


prajnoha:
Looks good, will merge. Thanks a lot!

In the tests, there's unused "do_alloc_test" - that was supposed to do all the tests like with the pre-buffered memory, but using the allocation instead, right?

bmarzins:
> In the tests, there's unused "do_alloc_test" - that was supposed to do all the tests like with the pre-buffered memory, but using the allocation instead, right?

Oops. Yeah. The comb_alloc_test* tests were supposed to test using malloc. I just called the wrong function in them. I'll fix that and update the pull request.

PR #124, opened 2022-03-16
Worker cleanup

### LINKED ISSUE #85: Event loop's epoll FD inherited from main process not closed in worker
Recently, we have resolved #33 which dealt with signal handling cleanup in worker after `fork` - that works fine now. But there's one more thing - even though we call the cleanup in the worker right at its initialization:

https://github.com/sid-project/sid/blob/8a5f96a1aab12c157dc04ceb6bcd5bc44fb199b2/src/resource/ubridge.c#L3950-L3951

...everything is cleaned up except the inherited `epoll` FD backing up the event loop in main process. Quick debug shows that we really call `sid_resource_destroy` with `sd_event_unref` inside which is called:

https://github.com/sid-project/sid/blob/8a5f96a1aab12c157dc04ceb6bcd5bc44fb199b2/src/resource/resource.c#L325-L326

But checking the `/proc/<worker_process_pid>/fd`, it shows the `epoll` FD is still there...

### LINKED ISSUE #108: worker template process
Of course, relying on the fact that we're not unwinding the call stack in the worker is far from ideal and it's definitely a pain spot with possible side effects, yes.

This one's a bit tricky though - we can't exit event loop in the parent process (we'd lose events/requests) and we can't exit event loop in the worker (at least not with `sd-event` loop which checks whether the PID hasn't changed on almost all API calls). 

For a clean solution here, we'd either need to change underlying event loop to something else than `sd-event` that supports cleaning up after forking (maybe `libev`?) OR we'd need to create a minimal template process for creating new workers.

(There's also a 3rd option - create our own event loop implementation where we won't be restricted, but we'd need to gain all the know-how that was put into the existing implementations.)

With the "worker template process", the main process would pass the request to this template process and the template process would then fork itself in turn, creating a new worker (without any call stack remnants). The template process would use a very simple event loop to accept these requests ("simple" meaning just to accept those specific requests and easy to exit in forked process).

I think the approach with the "worker template process" is more appropriate as with that, we're not restricting the event loop library we could possibly use. Also, starting a new worker process from a clean template seems much more straightforward. Disadvantage is that it adds intermediary between the main process and worker process, but only during creation. When the workers are already created, the communication can be conducted directly between the main process and that worker process.

_Originally posted by @prajnoha in https://github.com/sid-project/sid/issues/99#issuecomment-859083821_

bmarzins:
This patchset makes the worker process clean up the inherited resources before starting up the worker event loop.

Resolves #85, Resolves #108

### DIFF
--- a/src/resource/worker-control.c
+++ b/src/resource/worker-control.c
@@ -65,6 +65,9 @@ struct worker_control {
 	struct worker_init_cb_spec  init_cb_spec;
 	unsigned                    channel_spec_count;
 	struct worker_channel_spec *channel_specs;
+	/* The following members are for initialzing the worker after fork() */
+	char *                 worker_id;
+	struct worker_channel *worker_channels;
 };
 
 struct worker_channel {
@@ -682,7 +685,6 @@ int worker_control_get_new_worker(sid_resource_t *worker_control_res, struct wor
 	const char *            id;
 	char                    gen_id[32];
 	char **                 argv, **envp;
-	int                     r = -1;
 
 	*res_p = NULL;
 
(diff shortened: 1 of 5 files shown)


bmarzins:
@prajnoha, if this method looks o.k. to you, a possible next step would be to make the worker_control also save the argv and envp for external workers, so that they can also delay exec'ing until after the main process event loop has ended and all the resources have been cleaned up. That would keep them from exec'ing with the old file descriptors still open.

prajnoha (approved):
Haven't found or hit any issues with this approach so far, I think we can use it.

bmarzins:
Made external workers cleanup after forking as well.  The rebase was for a minor change to
[resource: worker-control: cleanup before starting worker](https://github.com/sid-project/sid/pull/124/commits/e9850914814f789fdafc94a6f42dfe331e6a6f27)
In the new version, I didn't delete the `*res_p = NULL;` line at the start of `worker_control_get_new_worker()`. The line isn't absolutely necessary, since `_get_worker()` already sets this to NULL before calling `worker_control_get_new_worker()`, but since callers check whether this is NULL to decide if they are in the main process or  a worker process, it seemed like the function should always explicitly set this value.

src/resource/worker-control.c
```diff
@@ -797,8 +797,7 @@ int worker_control_get_new_worker(sid_resource_t *worker_control_res, struct wor
 			                    worker_control->channel_spec_count) < 0)
 				goto out;
 
-			if (worker_control->init_cb_spec.cb)
```
prajnoha, line 798: Actually, just recalled, this provided a way for doing any preparation/cleanup right before the `exec` is called. There shouldn't be `NULL` dereference - the worker init fn is a property of `worker_control` which is inherited in the child process and still available before we call `exec`.

There's just one thing to care about - if we're preparing an external worker, the `init_cb_spec.cb` (the `_worker_init_fn` in our concrete case) is called with NULL for the worker resource: `_worker_init_fn(NULL, ...)`. So the `worker` resource won't be available there, but still we have a chance to do some preparation/cleanup before we do `exec`, if needed, even without the worker resource.

bmarzins: Oops. Yeah, I forgot that the external worker's `init_cb_spec.cb` callback wouldn't be `_worker_init_fn()`. I'll change the patch. Should I leave in the code to destroy all the resources starting from the top, under the assumption that we'll always want to do that, and the callback can copy-out any resources that it wants to keep. Or should we leave the destroying for the callback to do, like it does in `_worker_init_fn()`.

This does make a nice segue to something that has been bothering me. I assume the way that external workers will end up working is that SID will have multiple worker_control resources, each dedicated to a different kind of worker. All the plumbing to make SID pick the correct kind of worker is still work to be done then. This means that we can figure out how to have the sid main() function determine which worker spawned a new process later as well. It could be something as simple as a new `worker_control` bool called `active` which would be set after the `fork()`.

bmarzins:
In light of our discussions in #125, it seems like we probably want to leave any resource destroying to the callback handler, since its hard to know all the environments that an external worker could be called in. Also, it's possible that there will never be multiple worker_control resources at any one time, so there's no point in rushing to come up with a solution to differentiate between them. I'll clean up the external worker handling stuff a bit.  The channels cleanup work can be done as a separate pull request.

prajnoha:
OK, looks good, merging... Thanks!

PR #126, opened 2022-03-23
Channels fixes

bmarzins:
Here are fixes for issues I saw in the channels code, along with a fix from Peter.

### DIFF
--- a/src/resource/worker-control.c
+++ b/src/resource/worker-control.c
@@ -198,11 +198,11 @@ static int _create_channels(sid_resource_t *        worker_control_res,
 	return 0;
 fail:
 	while (i > 0) {
+		i--;
 		if (proxy_chans[i].fd >= 0)
 			close(proxy_chans[i].fd);
 		if (chans[i].fd >= 0)
 			close(chans[i].fd);
-		i--;
 	}
 
 	if (proxy_chans)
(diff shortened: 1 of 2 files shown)


src/resource/worker-control.c
```diff
@@ -531,7 +531,7 @@ static int
 				goto fail;
 			}
 
-			if (!is_worker && chan->spec->wire.ext.used && chan->spec->wire.ext.pipe.fd_redir >= 0) {
+			if (is_worker && chan->spec->wire.ext.used && chan->spec->wire.ext.pipe.fd_redir >= 0) {
```
prajnoha, line 534: Yeah, right... At first I was thinking about a possibility to be able to redirect output on main process side ("proxy") through a pipe to worker so worker can do some further processing on the data stream the proxy would normally just write (e.g. when writing a file by main process, then redirecting that writing to worker so it can do further reformatting, decorating the data with certain format or sending the data to different locations, files...). Still, we'd need the redirection on the worker side. If we ever needed this, we can easily extend the code, but probably not needed now - the use case seems quite artificial for now. The use case with redirecting the "pipe read end" on worker side to external worker's input is surely much more common and makes more sense as you say...

prajnoha (approved):
Looks good, will merge. Thanks for pointing out that redirection part!

PR #127, opened 2023-02-10
Misc minor fixes

bmarzins:
I looked through the buffer code to come up to speed on its usage tracking, and found some minor issues.


prajnoha:
OK, thanks a lot! Merging...

prajnoha:
(I've noticed the do_test_zero_add params were shuffled, so I've just added 4f33d4ba2fb7639c8bf8d8bf8583ab8627d4d9ea on top.)

Issue #128, opened 2023-02-17
Likely Bptree bug with prefix lookups


bmarzins:
I'm pretty sure I've spotted a bptree bug that can break _find() with LOOKUP_PREFIX, which could cause looking up a range of keys by prefix with the bptree_iter_* code to fail. But before I go about fixing it, I just want to make sure I'm not missing something here.

So, as far as I can see, the bptree code doesn't enforce that the keys in its internal nodes are actually present in the leaf nodes. If you have not done any deletions from the bptree, the key for the rightmost record in each leaf node will also exist in an internal node as the key for the pointer to the subtree for which it is the rightmost entry (except for the rightmost entry in the entire bptree, whose key is only in the leaf node). But if you delete the rightmost record from a leaf block, the code doesn't always remove its key from the internal nodes of the bptree. If you look in _delete_entry() for the case where the leaf block is still valid after the key is removed, the only function that is called is _remove_entry_from_node(), which never touches the tree outside that leaf node. After this, the key to the removed record would still be in an internal node somewhere (again, unless its the rightmost record in the entire bptree). I assume that you already know about this, since you reference count the keys. If the code guaranteed that when you removed an entry from the bptree, its key was removed from the internal nodes as well as the leaf node, then you wouldn't need to reference count the keys. Once an entry was removed, its key would always be safe to free. But having internal keys that don't exist in the leaf nodes could cause _find() with LOOKUP_PREFIX to fail.

Imagine we are looking for all entries with a key starting with Prefix_x.  These entries exist in two sibling leaf nodes. The first node (Leaf1) has Prefix_x:key_a as its rightmost member. The second node (Leaf2) has Prefix_x:key_b as its leftmost member. Since Prefix_x:key_a is the rightmost key in a leaf block, but not the rightmost member of its parent, its key will be in its parent's node (Node1). When looking for the first key whose prefix matches Prefix_x, the bptree _find() code will see the key in Node1, look in Leaf1, and find Prefix_x:key_a.

But if Prefix_x:key_a is deleted, the bptree code won't necessarily remove that key from Node1. After this deletion, when looking for the first key whose prefix matches Prefix_x, the code will still look in Leaf1, not Leaf2.  When it doesn't find any keys in the leaf that match Prefix_x, it will assume that there aren't any matching entries, because it looked in the leaf where they belonged and didn't find any.

[...]

PR #129, opened 2023-02-28
Bptree fixes and cleanups

bmarzins:
Aside from fixing some bptree bugs, the main thing that this MR does is to remove the unnecessary temporary memory allocations when splitting nodes and leaves, and to make sure that the bptree code does all the necessary memory allocations before ever modifying the tree.  This means that the code can't error out once it starts modifying the tree.  The MR also adds bptree unit tests, to make sure that everything works as expected.

### DIFF
--- a/src/internal/bptree.c
+++ b/src/internal/bptree.c
@@ -444,19 +444,6 @@ static void _destroy_node(bptree_t *bptree, bptree_node_t *n)
 }
 
 /*
- * Creates a new leaf by creating a node and then adapting it appropriately.
- */
-static bptree_node_t *_make_leaf(bptree_t *bptree)
-{
-	bptree_node_t *leaf;
-
-	if ((leaf = _make_node(bptree)))
-		leaf->is_leaf = true;
-
-	return leaf;
-}
-
-/*
  * Helper function used in insert_into_parent to find the index of the
  * parent's pointer to the node to the left of the key to be inserted.
  */
@@ -505,10 +492,11 @@ static bptree_node_t *
 	bptree_key_t  *new_bkey;
 	int            insertion_index, split, i, j;
 
-	if (!(new_leaf = _make_leaf(bptree)))
+	if (!(new_leaf = _make_node(bptree)))
 		return NULL;
+	new_leaf->is_leaf = true;
 
-	insertion_index = 0;
+	insertion_index   = 0;
 	while (insertion_index < bptree->order - 1 && strcmp(leaf->bkeys[insertion_index]->key, bkey->key) <= 0)
 		insertion_index++;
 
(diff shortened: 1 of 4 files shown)


bmarzins:
Resolves #128

bmarzins:
I didn't change the function returns from the bptree code, although a number of function's returns don't make much sense anymore and could be cleaned up. For one thing, many are a vestige of the original code, which didn't have a structure to store the root node, so inserts and removes had to return the current root, because it could have changed. For another thing, a number of functions no longer need to have the ability to return failure, since they no longer have a failure path.

prajnoha:
> Aside from fixing some bptree bugs, the main thing that this MR does is to remove the unnecessary temporary memory allocations when splitting nodes and leaves, and to make sure that the bptree code does all the necessary memory allocations before ever modifying the tree. 

Great, looks good, thank you!

So, now we are sure that we don't end up with a messed up tree when modifying a single key. Now the question is if we could extend that further to include a set of changes - practically creating a simple transaction. Normally, I'd put such code at the "kv-store" level that would monitor all the operations going down to the backend (bptree), keeping a journal of updates done so far, checking the results from the backend and if any of the operations failed, revert all the changes recorded in the journal. It depends on how much extra work would that encompass to add such a functionality, I haven't thought about details yet...

> I didn't change the function returns from the bptree code, although a number of function's returns don't make much sense anymore and could be cleaned up. 

Sure, we can clean that up later, it doesn't matter now.

> Also, the asserts in internal/bptree.c were mostly there as sanity checks while testing it. I can remove them if you want.

I think we can keep them in - if anything is wrong there, we'll see that right away at leass. Actually, maybe, we could start adding more asserts in other part of the code where applicable.

bmarzins:
> So, now we are sure that we don't end up with a messed up tree when modifying a single key. Now the question is if we could extend that further to include a set of changes - practically creating a simple transaction. Normally, I'd put such code at the "kv-store" level that would monitor all the operations going down to the backend (bptree), keeping a journal of updates done so far, checking the results from the backend and if any of the operations failed, revert all the changes recorded in the journal. It depends on how much extra work would that encompass to add such a functionality, I haven't thought about details yet...

Yeah. I'm assuming that since we can no longer mess the tree up during changes, we will just be able to track the changes and revert them on failure, but I haven't looked into that code yet. The other alternative is something like a copy-on-write btree, but the problem with that is that copying a leaf block usually means copying all the blocks that point to it, and then the blocks that point to them, up till the root.  But if you have leaf block chaining, you could end up needing to copy all the leaf blocks, an thus duplicate the whole tree for every transaction. This seems like a lot of work to deal with a not very likely error case. There are ways around that issue, but they add complexity too, so I'm hoping to just keep a list of changes that need to be reverted.

PR #130, opened 2023-04-13
Add ability for kv store sync to rollback to previous state on failure.

bmarzins:
_sync_main_kv_store() can now rollback if it encounters a failure syncing the kv store. Since unsetting an entry cannot fail, it delays performing these actions until it has already performed the ones that can fail.  For actions that can fail (setting, adding and removing), the kv callbacks now have access to an opaque pointer to the current kv_store_value.  They save this value instead of destroying it. If all the the actions succeed, the saved values are destroyed.  Otherwise, the saved values are used to rollback to state of the database. The new value is destroyed and depending on whether the saved value was NULL or not, the entry is either removed, or it's value is replaced by the saved value. Neither of these actions can fail.  It is possible that the old value was saved, but the entry was never updated.  In this case, the new and saved values will be the same during rollback, and nothing is done. The end result is that it is always possible to revert the state of the database while only calling functions that do not fail.

There are unit tests for this code, but they trigger the rollbacks by writing an invalid entry to worker kv store. This means that the code to deal with failures in updating the entry is not tested.  It's probably possible with some fancy mocking of the allocation code.

### DIFF
--- a/src/base/buffer-type-linear.c
+++ b/src/base/buffer-type-linear.c
@@ -294,20 +294,20 @@ static bool _buffer_linear_is_complete(struct sid_buffer *buf, int *ret_code)
 	return result;
 }
 
-static int _buffer_linear_get_data(struct sid_buffer *buf, const void **data, size_t *data_size)
+static int _buffer_linear_get_data(struct sid_buffer *buf, size_t pos, const void **data, size_t *data_size)
 {
 	switch (buf->stat.spec.mode) {
 		case SID_BUFFER_MODE_PLAIN:
 			if (data)
-				*data = buf->mem;
+				*data = buf->mem + pos;
 			if (data_size)
-				*data_size = buf->stat.usage.used;
+				*data_size = buf->stat.usage.used - pos;
 			break;
 		case SID_BUFFER_MODE_SIZE_PREFIX:
 			if (data)
-				*data = buf->mem + SID_BUFFER_SIZE_PREFIX_LEN;
+				*data = buf->mem + SID_BUFFER_SIZE_PREFIX_LEN + pos;
 			if (data_size)
-				*data_size = (buf->stat.usage.used) ? buf->stat.usage.used - SID_BUFFER_SIZE_PREFIX_LEN : 0;
+				*data_size = (buf->stat.usage.used) ? buf->stat.usage.used - SID_BUFFER_SIZE_PREFIX_LEN - pos : 0;
 			break;
 		default:
 			return -ENOTSUP;
--- a/src/base/buffer-type-vector.c
+++ b/src/base/buffer-type-vector.c
@@ -278,20 +278,20 @@ static bool _buffer_vector_is_complete(struct sid_buffer *buf, int *ret_code)
 	return true;
 }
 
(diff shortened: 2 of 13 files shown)


prajnoha:
Great, thanks a lot for this, looks good.

Now, I'm just thinking whether it would be even possible to hide the "rollback" and "unset" buffer inside the kv-store layer, so we'd end up with something like this:

```
kv_store_transaction_begin 
   /* ...kv store update/unset calls */
if failure
   then kv_store_transaction_rollback
else
   kv_store_transaction_end
```

Where `kv_store_transaction_begin` would create the rollback buffer, returning a transaction handle (which would be assigned to the kv-store inside). Maybe also locking the whole kv_store for all other KV store calls without this handle, that is, not allowing any other interleaved operations that do not have this exact transation handle at hand (but that would require all the existing kv-store API functions to accept the handle, so we probably don't need to go that far).

The `kv_store_transaction_rollback` would do the iteration over the rollback buffer and do the rollback record by record . Similarly for `kv_store_transaction_end` that would instead discard the rollback buffer content. So exactly the same as you do now at the end of `_sync_main_kv_store`, just it would happen inside kv-store layer.

(In that case, the `_update_fn` would still return either "0" or "1" and the functionality of "return 2" would be inferred automatically inside kv-store if we had "1" and at the same time we're under an active transaction.)

That way, we wouldn't need to expose the (even though opaque) `struct kv_store_value` and `destroy_kv_store_value`. Also, if we needed to reuse the kv-store with rollback functionality for something else, we wouldn't need to create/copy the rollback and unset buffer/add to the buffer/call the rollback or destroy_kv_store_value - this functionality would already be packaged inside the `kv-store` resource itself and ready for reuse.

Just a thought, maybe there are some practical obstacles in such a model that I haven't included... What do you think?

bmarzins:
I think your suggestion can work. I do have a question about how you see it working. I'm not sure that we need to pass back a handle for anything. Assuming that we would want to sleep and let other code run between when we start and end a transaction, we just need in-transaction versions of kv_store_set_value() and kv_store_unset(). They don't need to pass in the handle. the kv-store code just needs to make sure that only the in-transaction versions will succeed while in a transaction.

This seems similar to how locking works, where you don't get a handle to tell you that you have the lock. You know because grabbing the lock succeeded.  I can see the possibility of needing the ability to query the kv-store to see if it's in a transaction, or needing the ability to either start a transaction or add a callback to the event loop which gets run when the current transaction ends. Am I missing some need for the handle here?

bmarzins:
Here's V2, where I pulled the transaction logic and buffers into the kv-store.

prajnoha:
> I think your suggestion can work. I do have a question about how you see it working. I'm not sure that we need to pass back a handle for anything. Assuming that we would want to sleep and let other code run between when we start and end a transaction, we just need in-transaction versions of kv_store_set_value() and kv_store_unset(). They don't need to pass in the handle. the kv-store code just needs to make sure that only the in-transaction versions will succeed while in a transaction.

Yeah, you're right. There's actually no need for the extra handle - we have everything we need in the `struct kv_store` to keep the needed state. I was just flushing out the idea quickly without thinking details :)

The V2 looks really good, the code got also simplified, I think. Thanks a lot for the patches! I'll merge these...

PR #132, opened 2023-05-11
Fix variable alignment issues.

bmarzins:
These are all the alignment issues I found while looking through the code.

### DIFF
--- a/src/iface/iface.c
+++ b/src/iface/iface.c
@@ -76,31 +76,36 @@ void sid_result_free(struct sid_result *res)
 int sid_result_status(struct sid_result *res, uint64_t *status)
 {
 	size_t                       size;
-	const struct sid_msg_header *hdr;
+	const struct sid_msg_header *hdr_p;
+	struct sid_msg_header        hdr;
 
 	if (!res || !status)
 		return -EINVAL;
-	sid_buffer_get_data(res->buf, (const void **) &hdr, &size);
-	*status = hdr->status;
+	sid_buffer_get_data(res->buf, (const void **) &hdr_p, &size);
+	memcpy(&hdr, hdr_p, sizeof(struct sid_msg_header));
+	*status = hdr.status;
 	return 0;
 }
 
 int sid_result_protocol(struct sid_result *res, uint8_t *prot)
 {
 	size_t                       size;
-	const struct sid_msg_header *hdr;
+	const struct sid_msg_header *hdr_p;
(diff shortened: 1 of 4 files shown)


prajnoha:
Thanks! I just have one question about this part:

```
struct kv_store_value {
  ...
  char                       data[] __attribute__((aligned));
};
```

Will compiler deduce we need pointer alignment or do we need to state that directly with `aligned(sizeof(intpr_t))` or so?

bmarzins:
> Thanks! I just have one question about this part:
> 
> ```
> struct kv_store_value {
>   ...
>   char                       data[] __attribute__((aligned));
> };
> ```
> 
> Will compiler deduce (from the `[]`) we need pointer alignment or do we need to state that directly with `aligned(sizeof(intpr_t))` or so?

You need that `__attribute__((aligned))` to tell the compiler that data needs to start at pointer alignment. Otherwise, since it's an array of chars, it just needs char alignment, which means it can start anywhere.
```
[bmarzins@octiron putzing]$ cat align.c
#include <stddef.h>
#include <stdio.h>

struct test {
	short int pad;
	char data[];
};

struct aligned_test {
	short int pad;
	char data[] __attribute__((aligned));
};

int main(void)
{
	printf("test data offset: %lu\n", offsetof(struct test, data));
	printf("aligned test data offset: %lu\n", offsetof(struct aligned_test, data));
	return 0;
}	
[bmarzins@octiron putzing]$ ./align 
test data offset: 2
aligned test data offset: 16
```

bmarzins:
So `__attribute__((aligned))` is supposed to give the maximum useful alignment for your system.  I don't know why it gave me a 16 byte alignment instead of an 8 byte one (dunno what could possibly need more than 8 byte alignment on a 64 bit system), but replacing it with `__attribute__((aligned (sizeof(intptr_t))))` gives me the expected pointer alignment

```
test data offset: 2
aligned test data offset: 8
```
I can change that and update the MR if you want.

prajnoha:
> So `__attribute__((aligned))` is supposed to give the maximum useful alignment for your system. I don't know why it gave me a 16 byte alignment instead of an 8 byte one (dunno what could possibly need more than 8 byte alignment on a 64 bit system)

Hmm, maybe some specialized CPU registers are taken into account (the 128-bit vector ones/SSE...)? Just guessing...

> I can change that and update the MR if you want.

...that's OK, let's keep what compiler think is the best for the architecture. Will merge then. Thanks!

PR #147, opened 2023-09-26
Misc cleanups

bmarzins:
I've gotten distracted with other things, so I haven't made much progress on the remove work yet, but here are some fixes that I've accumulated while looking through the code to understand what needs to be done.  I'm not quite sure if the final commit is the correct solution (erroring out instead of just skipping the failed sysfs attribute and going on). But the most likely cause of this seems to be that the sysfs device has disappeared. Right now this happens during remove events, but even when those have a different call path, the sysfs device could still get removed while we are processing an earlier uevent. in this case, I don't see much point in continuing to update the database for the device. But if you'd rather the code skip adding that attribute and continue with the function, I can change it.

### DIFF
--- a/src/iface/service-link.c
+++ b/src/iface/service-link.c
@@ -70,16 +70,14 @@ void service_link_destroy(struct service_link *sl)
 	free(sl);
 }
 
-int service_link_add_notification(struct service_link *sl, service_link_notification_t notification)
+void service_link_add_notification(struct service_link *sl, service_link_notification_t notification)
 {
 	sl->notification |= notification;
-	return 0;
 }
 
-int service_link_remove_notification(struct service_link *sl, service_link_notification_t notification)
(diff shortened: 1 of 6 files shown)


prajnoha:
> ...the sysfs device could still get removed while we are processing an earlier uevent. in this case, I don't see much point in continuing to update the database for the device. But if you'd rather the code skip adding that attribute and continue with the function, I can change it.

Yup, right now, we should skip. I'm just fiddling with the "ready" state for devices, so later on, we will also cover this situation (when we expect certain part/attribute to exist on the system, but it's not there, obviously because the device has been removed in the meantime) by marking the device with proper state and erroring out, giving a chance for SID to react somehow (expecting a subsequent event to be "REMOVE" or restart the operation as if it was "REMOVE" etc.).

Thanks for the patches!

PR #148, opened 2023-10-09
More bugfixes and remove infrastructure

bmarzins:
I still keep having other things on my plate that are keeping me from being able to finish up the remove work. Instead, I'll just send you what I currently have, so you can see the direction I'm heading. I've split out the remove work from the regular scan work, although it still uses the "usid scan" command, since there is a lot of code overlap with regular scans. It currently doesn't do anything special on removes. That's what I'm working on next. Also, here are fixes for two bugs that were keeping resyncing the device hierarchy from working on removes.


prajnoha:
Yup, that looks good. The "fix freed memory access in _kv_delta_set" - wow, that's was quite a bug, slipped through the cracks. Thanks!

